Issue #18, opened 2023-11-19
Country Zoom


SeanKennyNF:
From the bug report spreadsheet,

"When a user selects a country in the filters, the map automatically zooms in on that country"

PR #19, opened 2023-11-19
Country Zoom

SeanKennyNF:
Implements #18 

Screenshot010


contexts/arbo-context.tsx
```diff
@@ -84,6 +93,28 @@ export const arboReducer = (state: ArboStateType, action: ArboAction) => {
...
+const adjustMapPositionIfCountryFilterHasChanged = (
+  action: ArboAction
+): void => {
+  if (action.payload.filter === "country") {
+    const allSelectedCountryBoundingBoxes = action.payload.value
+      .map((countryName: string) => getBoundingBoxFromCountryName(countryName))
+      .filter((boundingBox: CountryBoundingBox) => !!boundingBox);
+
+    if(allSelectedCountryBoundingBoxes.length === 0) {
+      action.payload.map.fitBounds([-180, -90, 180, 90]);
```
SeanKennyNF, line 105: Resets the position of the map to the equator + prime meridian if you deselect the country you have selected or if you've only selected countries that don't exist in `countryBoundingBoxData` (currently every country in pathotracker has been added to `countryBoundingBoxData` so there shouldn't be any problems but pointing it out just in case.

lib/country-bounding-boxes.ts
```diff
@@ -0,0 +1,217 @@
+export type CountryBoundingBox = [number, number, number, number];
+
+const countryBoundingBoxData: Record<string, CountryBoundingBox | undefined> = {
```
SeanKennyNF, line 3: So most of this was pulled from this person who made some creative commons bounding boxes for countries.
https://github.com/sandstrom/country-bounding-boxes

I had to make some adjustments because there was some countries missing in that repo and some of the names in that repo didn't match the names we use in the web app Ex. "United States" vs "USA".

I also had to make some adjustments for countries that cross the 180 longitude line (Russia + Fiji) because the map would just reset to the prime meridian + equator if one of those countries was selected so I just cut off the tiny part of their countries that cross that line, I don't think it's super noticeable. I left Antarctica alone.

Basically this commit is a summary of everywhere I deviated from the publicly available dataset. When data wasn't available I just went to Google Earth and eyeballed it. I tested out all of the countries I eyeballed and I think I did an okay job. https://github.com/PathoTracker/Pathotracker/pull/19/commits/a7c53b7b2191efec70c8894a2a9accedf941cfa1

The map resets it's position if you select two countries that are really far away because it can't fit both of them on the same screen or if you pick two countries really close to longitude 180 and on either side (Ex. Fiji and USA) so there's some stuff we can iterate on here but at least this gives us something to iterate on hahaha

lib/country-bounding-boxes.ts
```diff
@@ -0,0 +1,217 @@
...
+  ];
+};
+
+export const combineCountryBoundingBoxes = (
+  boundingBoxes: CountryBoundingBox[]
+): CountryBoundingBox => {
+  return boundingBoxes.reduce((accumulator, currentValue) =>
+    combineTwoCountryBoundingBoxes(accumulator, currentValue)
+  );
+};
```
SeanKennyNF, line 217: All of this code here is so that when you pick more than one country it tries it's best to fit both of them in one box. If the box is too big it just resets the position of the map.

lib/country-bounding-boxes.ts
HimanshuRanka: This is awesome!

HimanshuRanka (approved):
This is really great! Thanks for putting in the work to get this going

Just some things to note:
- We would want to eventually add country highlighting to show which countries are selected.

LGTM otherwise, will iterate on this

Issue #33, opened 2023-11-20
Update the Log to link to the Landing Page


HimanshuRanka:
The link in the top left corner links to https://pathotracker.vercel.app/Explore which is a dead link and just 404s the user. If you're on a deeper nested page, it just replaces the last term in your path with "Explore" Ex. https://pathotracker.vercel.app/pathogen/arbovirus/dashboard becomes https://pathotracker.vercel.app/pathogen/arbovirus/Explore

Issue #32, opened 2023-11-20
Antibody Filter does not work as expected


HimanshuRanka:
The way the antibody information is pulled from the backend is different from how all other information is pulled (specifically antibody is a multiselect field of data) this causes inconsistencies with the data filtering and analytics mechanisms. This is because selecting a specific antibody means you need to filter for all studies that have that antibody irrespective of whether that is the only antibody or there are others (this filtering mechanism also needs to be worked on and improved)

Fix the antibodies filter to function as we would expect it to on all parts of the webpage

Issue #29, opened 2023-11-20
Add LINK button to arbo table


HimanshuRanka:
The link button should take a user to the url provided for that estimate. Each row on the data table should have a link button. 

Alternatively would a better appraoch be to highlight some unique field that is clickable instead of needing a button? Consider the same - Check with Harriet and Mairead on what the users will find most useful

HimanshuRanka:
For now we went with the approach of having a LINK button. The text in the link button is the domain of the URL

SeanKennyNF:
I believe addressed in https://github.com/PathoTracker/Pathotracker/pull/36. Going to close if that's alright!

PR #36, opened 2023-11-22
Antibody filter

HimanshuRanka:
This PR resolves: #33 #29 #32 and then some based on cosmetic things I believe could be improved

- Added a Link button to the Arbotracker table that takes the user to the publication from which said estimate was extracted
- Updated the style of some parts of the table so that it is easier to understand the data and the decimals are not super long
- Updated the logo to link to the landing page
- Improved the styling of the filters to look cleaner and work more intuitively
- Fixed the issue with the data resetting to all data when the filtered data was empty (this was not intuitive and so now the map and visualizations are blank when the filters amount to no data
- Fixed the antibody filter such that it now works for both the map and the data table and successfully filters for antibodies.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -5,6 +5,7 @@
 @layer base {
   :root {
     --background: 206 28% 29%;
+    --background-hover: 206 28% 43%;
     --foreground: 0 0% 100%;
 
     --card: 0 0% 100%;
--- a/app/pathogen/arbovirus/analyze/ArboDataTable.tsx
+++ b/app/pathogen/arbovirus/analyze/ArboDataTable.tsx
@@ -10,7 +10,7 @@ export default function ArboDataTable() {
   const dataQuery = useArboData();
   const state = useContext(ArboContext);
   if(state.filteredData?.length > 0 || (dataQuery.isSuccess && dataQuery.data)) {
-    return <DataTable columns={columns} data={state.filteredData} />;
+  return <DataTable columns={columns} data={state.filteredData?.length <= 0 && !state.dataFiltered ? dataQuery.data.records : state.filteredData} />;
   } else {
     return <>Loading Data ...</>
   }
--- a/app/pathogen/arbovirus/analyze/ArboDataTable.tsx
+++ b/app/pathogen/arbovirus/analyze/ArboDataTable.tsx
@@ -10,7 +10,7 @@ export default function ArboDataTable() {
   const dataQuery = useArboData();
   const state = useContext(ArboContext);
   if(state.filteredData?.length > 0 || (dataQuery.isSuccess && dataQuery.data)) {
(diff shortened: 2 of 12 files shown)


app/globals.css
```diff
@@ -5,6 +5,7 @@
 @layer base {
   :root {
     --background: 206 28% 29%;
+    --background-hover: 206 28% 43%;
```
SeanKennyNF, line 8: Commenting on random line for thread because it's a bit unclear to me where this is coming from but I'm experiencing a possible regression here?

When I visit Arbotracker for the first time the visualizations and tables don't seem to be filled in until I select a filter. I'm guessing this PR changes the state management a bit and that's causing this?
peek2222

SeanKennyNF (approved):
Cool! I've got a QQ about some behavior that seems to have changed here. Some of the finer details about the state management are lost on me so I'm not sure _exactly_ what caused it but probably worth fixing? Table looks nice though :)

app/globals.css
SeanKennyNF: Actually maybe it's the change in `app/pathogen/arbovirus/analyze/ArboDataTable.tsx` :eyes:

The state not updating until a filter is applied and re-applied seems like maybe that ternary was keeping whatever underlying state management problem was happening here under control?

HimanshuRanka:
Oooo yeah you’re right! I’ll fix that tomorrow and push it back up

HimanshuRanka:
Made a cahnge but have an inkling that there may be a better way to do the same thing I am doing above (if we have 20 visualizations this will lead to a lot of repetitive code. So extract to a hook? or is there something even simpler?)

SeanKennyNF (approved):
Yeah so I tried my hand at this. I feel like the simpler way here is to consolidate the `dataQuery` and `filteredData` into one variable. The query finishing would initialize it and then any filters you put on top would manipulate that same state variable. I made a branch with my changes and it looks like everything works okay.

I made a commit here that you hopefully find simpler than the ternary? It's a bit longer but you won't have to repeat that whole ternary anywhere if you set up another display because the context provider is the only one that needs access to the raw data from the query after my commit and there shouldn't be any more state problems?

Anyways hopefully you like this and also agree that it makes the state management a bit simpler. If so, let's pull it into here and then we'll be off! If not, totally cool and this still looks good to me :)

5c6efec5c8408bffa8a406ff129320ef807e3238

HimanshuRanka:
Looks great! Merged it in. Just to confirm I understand when we  get the site mounted and the context initializes, we use the useEffect to check if the filtered data is empty or not and populate it with the prefetched data if not?

SeanKennyNF:
> Looks great! Merged it in. Just to confirm I understand when we get the site mounted and the context initializes, we use the useEffect to check if the filtered data is empty or not and populate it with the prefetched data if not?

Yeah exactly :+1:

PR #40, opened 2023-11-26
[DO NOT MERGE] Draft PR for table enhancements. 

EliasPinno:


gurickkooner (approved):

components/ui/data-table.tsx
```diff
@@ -1,5 +1,5 @@
 "use client";
-
+// Adding a comment to the data table.
```
HimanshuRanka, line 2: Can we remove this comment?

components/ui/data-table.tsx
```diff
@@ -61,6 +63,19 @@ export function DataTable<TData, TValue>({
     },
   });
 
+  const [selectAll, setSelectAll] = useState(false);
```
HimanshuRanka, line 66: initial state should be true as all columns are selected. 
image

HimanshuRanka: Also the state should update when something is unselected manually:

image

components/ui/data-table.tsx
```diff
@@ -81,6 +96,14 @@ export function DataTable<TData, TValue>({
             </Button>
           </DropdownMenuTrigger>
           <DropdownMenuContent align="end" className="bg-foreground">
+          <DropdownMenuCheckboxItem
```
HimanshuRanka, line 99: Instead of having the "All" Be an option in dropdown could we make it a Button at the top. Additionally add in a clear button at the top as well so that a user can select all or clear all with a single click. 

Specifically have the buttons be "Select All" and "Clear All". 

- Disable the select all button when all are selected and Disable the Clear All button when None are selected. 

This way we will not need to use state in the way you have above.

HimanshuRanka: Reasoning: https://ux.stackexchange.com/questions/43672/should-multi-select-dropdown-box-have-an-all-option

I think interacting with the All as a select option just feels like bad UX to me now so I feel the above approach may be better (something we will find out more in user testing) 

You could have the buttons be top and bottom or side by side, can make a call on that as needed.

HimanshuRanka (changes requested):
Left a couple comments to change the UX of the functionality!

components/ui/data-table.tsx
EliasPinno: I agree completely, and implemented the requested changes. See the gif at the top of the PR for details.

components/ui/data-table.tsx
EliasPinno: Done.

Issue #25, opened 2023-11-20
Update Arbo Filter Bar


HimanshuRanka:
Pending discussion:

Filters needed: 
* PENDING

Update the arbotracker filter bar to 
* have informative tool tips
* Be subdivided by category of filtler
* Have a reset filters button

harrietware:
Please add a filter for WHO region

PR #41, opened 2023-11-26
feature: added filter section headings, fixed some filterbox styling,…

gurickkooner:
Issue #25 Update Arbo Filter Bar
-Added section headings grouping together the filter fields (created custom component in components/customs)
-Added a tooltip used fontawesome icons (npm installed and added to the SectionHeader.tsx component code)
-Added reset filter button (changes to  filter.tsx which is the button and reset filter function with a newly created actions which results in changes to arbo-context to handle the reset filter action)


app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -162,11 +202,19 @@ export default function Filters(props: { map?: mapboxgl.Map | null }) {
...
                 filters.data["pathogen"],
                 map,
-                data ? data.records : [],
+                data ? data.records : []
               )}
             </div>
           )}
         </div>
+        <div>
+          <button
```
HimanshuRanka, line 211: Can we make the button sticky with the bottom of the filter bar? so that the button is always at the bottom no matter what.

app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -29,64 +53,77 @@ const addFilterMulti = (
...
     return (
       <div>
         <div className="p-0">
-          {/*<div>*/}
-          {/*    <SectionHeader*/}
-          {/*        header_text={"Demographics"}*/}
-          {/*        tooltip_text={"Participant related data"}*/}
-          {/*    />*/}
-          {/*</div>*/}
+          <div>
```
HimanshuRanka, line 111: Can we make the filter bar scrollable so that filters do not get cut off when the screen size is small

image

components/customs/SectionHeader.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { useState } from "react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faInfoCircle } from "@fortawesome/free-solid-svg-icons";
```
HimanshuRanka, line 3: replace with lucide

components/customs/SectionHeader.tsx
```diff
@@ -0,0 +1,39 @@
...
+interface SectionHeaderProps {
+  header_text: string;
+  tooltip_text: string;
+}
+
+const SectionHeader: React.FC<SectionHeaderProps> = ({
+  header_text,
+  tooltip_text,
+}) => {
+  const [showTooltip, setShowTooltip] = useState(false);
```
HimanshuRanka, line 15: Can we use the shadcn tooltip here? https://ui.shadcn.com/docs/components/tooltip

Reasoning: Shadcn and other ui libraries (headless, material, chakra, mantine) etc. all have the benefit of having a lot of small things preset. Accesssibility, styling, themeing etc. So we want to use them as much as possible. Always look at the shadcn website to check if there exists a componenent

I specifically really like shadcn because you can customize the components top to bottom which gives us full control over the functionality etc and we do not need to create wrappers. Similar feeling for Headless UI.

Shadcn itself is built on radix so that would be the second place to check.

contexts/arbo-context.tsx
```diff
@@ -105,14 +106,21 @@ export const arboReducer = (state: ArboStateType, action: ArboAction) => {
...
         ...state,
         filteredData: action.payload.data,
         selectedFilters: initialState.selectedFilters,
-        dataFiltered: false
+        dataFiltered: false,
+      };
+    case ArboActionType.RESET_FILTERS:
+      return {
+        ...state,
+        filteredData: [],
```
HimanshuRanka, line 121: I do not want to hardcode this logic here. Instead we should call the filterData function but have the selected filters be an empty list.

In fact we do not NEED another dispatch action as this basically does what filtered data would do if we manually cleared all filters, so let us keep that logic consistent

hooks/useMap.ts
```diff
@@ -135,6 +139,7 @@ export const getMapboxLatitudeOffset = (map: mapboxgl.Map | undefined) => {
   return 0
 }
 
+//creates map on click listeners to handle opening pop ups
```
HimanshuRanka, line 142: This is a useful comment +1

hooks/useMap.ts
```diff
@@ -182,6 +187,7 @@ function initializeMap(map: mapboxgl.Map, data: any, pathogen: "Arbovirus" | "Sa
   );
 }
 
+//Hook called in MapAndFilters to initalize mapbox map, content, logic
```
HimanshuRanka, line 190: I like the description, but the whole point of being reusable is we do not hardcode WHERE it is called so remove that (in the situation we call it elsewhere this comment becomes obsolete. that is the exact problem we want to avoid)

contexts/arbo-context.tsx
gurickkooner: I have changed this to call the filterData function as mentioned above it will now be as follows (in the next push):
 case ArboActionType.RESET_FILTERS:
      return {
        ...state,
        filteredData: filterData(action.payload.data, {}),
        selectedFilters: initialState.selectedFilters,
        dataFiltered: false,
      };

hooks/useMap.ts
gurickkooner: Understood, removed the comment.

components/customs/SectionHeader.tsx
gurickkooner: I have changed it to use shadcn and it works with the proper tool tip text showing up. I have encountered some issues attempting to change the styling such as the background color and width are not being applied. I did use the manual way to setup with the src code places inside the components/ui folder. I am currently unsure why the styling is not being applied.

app/pathogen/arbovirus/dashboard/(map)/MapAndFilters.tsx
```diff
@@ -1,3 +1,10 @@
+/**
```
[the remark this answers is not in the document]
gurickkooner: Have removed the scattered comments

components/customs/SectionHeader.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { useState } from "react";
```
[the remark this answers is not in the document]
gurickkooner: I have applied the pb-4 padding as shown above. I am currently having an issue with applying changes to the shadcn tooltip for the width and background color as they are not being applied to the pop over component.

app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -1,3 +1,25 @@
+/**
+ * @file Filters Component
+ * @description This component renders a set of filters for the Arboviruses dashboard.
+ * It includes dropdowns for age group, sex, country, assay, producer, sample frame, antibody, and pathogen.
+ * The filters are dynamically updated based on user interactions and are synchronized with the global state.
+ * The component fetches arbovirus data and filter options from the API, rendering the filters once the data is available.
+ * It uses the ArboContext to manage global state and interacts with the map using the mapboxgl library.
+ *
+ * Directory Structure:
```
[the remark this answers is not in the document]
gurickkooner: Have removed it

app/pathogen/arbovirus/dashboard/filters.tsx
gurickkooner: Screenshot 2023-12-01 at 16-12-13 Pathotracker

gurickkooner: added a scrollbar so that in smaller screen the user can scroll down

SeanKennyNF:
Hey @gurickkooner!
The PR I just merged had some merge conflicts with yours so I went ahead and fixed them on your PR.
Sorry for the trouble! I think I did it correctly but if you see anything weird please let me know :heart:

app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -136,29 +154,120 @@ export default function Filters({ excludedFields = [] }: FiltersProps) {
...
-        />
-        <FilterSection
-          headerText="Study Information"
-          headerTooltipText="Filter on different types of study based metadata"
-          fields={studyInformationFilters}
-          state={state}
-          filters={filters}
-          data={data}
-          filterableFieldToLabelMap={filterableFieldToLabelMap}
-        />
```
SeanKennyNF, line 161: Heyo, I'm down for anything tbh but I would consider what I've done with `<FilterSection>` vs what's here. The 15 lines deleted here are equivalent to the 100 lines below so I would consider switching back to `<FilterSection>`

app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -136,29 +154,120 @@ export default function Filters({ excludedFields = [] }: FiltersProps) {
...
+            )}
+          </div>
+          <div>
+            {buildFilterDropdown(
+              "pathogen",
+              "Pathogen",
+              state,
+              filters.data["pathogen"],
+              data ? data.records : []
+            )}
```
SeanKennyNF, line 260: Another reason to switch back to the `<FilterSection>` approach is that you will break this functionality from my PR if you delete my `<FilterSection>` stuff.

See discussion https://github.com/PathoTracker/Pathotracker/pull/54#discussion_r1411088344

package.json
```diff
@@ -9,6 +9,8 @@
     "lint": "next lint"
   },
   "dependencies": {
+    "@fortawesome/free-solid-svg-icons": "^6.4.2",
+    "@fortawesome/react-fontawesome": "^0.2.0",
```
SeanKennyNF, line 13: Unused dependencies or am I missing something?

components/customs/SectionHeader.tsx
```diff
@@ -0,0 +1,64 @@
...
+            <TooltipTrigger>
+              <div
+                onMouseEnter={() => setShowTooltip(true)}
+                onMouseLeave={() => setShowTooltip(false)}
+                className="h-5 w-5 text-gray-500 cursor-pointer"
+              >
+                &#9432;
+              </div>
+            </TooltipTrigger>
+            {/* Tooltip Content */}
```
SeanKennyNF, line 39: The tooltip isn't triggering for me on Firefox or Chrome. I wonder if it works for you if you've done something browser speicific or if one of the merge conflicts somehow broke this functionality?

SeanKennyNF (commented):
Heyo! I've just got a couple suggestions and a place where I think a merge conflict wasn't sorted out correctly. I'd be super down to pair program on this, I admit that I have a role in the whole mess with the filter heading code hahahaha

package.json
gurickkooner: Yeah that is unused I will remove since ended up using shadcn

components/customs/SectionHeader.tsx
gurickkooner: I am using firefox on linux and it works. I think the offset of where to hover is a bit tricky to place. Perhaps can look into making it a click to open instead of a hover.

gurickkooner: Will test again on chrome to make sure. Not sure at the moment why it would not work.

SeanKennyNF: Oh sorry just to clarify my testing was done on a Windows machine.

SeanKennyNF: Discussed, I wasn't waiting long enough hahaha

app/pathogen/arbovirus/dashboard/filters.tsx
gurickkooner: Does this affect anything on the serotracker dashboard? after I made the changes it seemed to cause some error loading the serotracker dash. Not sure could have been something else.

gurickkooner: Have added it back sorry about that did not analyze it properly the first time. I agree this makes it so much cleaner thanks.

SeanKennyNF (approved):
Makes sense to me, we can hit any tiny things on the next PR you have coming anyways :)

Maybe in the next PR we could rename `header_text` and `tooltip_text` to `headerText` and `tooltipText` for consistency?

HimanshuRanka (approved):

PR #54, opened 2023-11-30
Switch to using `react-map-gl` instead of `mapbox-gl` (Resolves #35)

### LINKED ISSUE #35: Filters are not persisting between map and analysis
We need to get the filters to persist across both pages.

This is probable a breakdown on the useMap hook side. When the map gets painted on, the source is not updating correctly/with the correct data/ at all. We need to somehow trigger that update.

SeanKennyNF:
Resolves #35 

I left a comment at the top of my changed files that summarizes why I think switching to `react-map-gl` is a good idea despite the temporary grossness it introduces and why switching to this library actually resolves #35 in a way that's probably cleaner than what I would have had to do without this library.

Again, comment at the top of the changed files for more details, feel free to check out the vercel preview if there's anything you want to make sure I didn't break :)

peek-29-11-2023


components/ui/pathogen-map/pathogen-map-popup.tsx
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  if (popUpInfo.visible === false) {
+    return null;
+  }
+
+  // Not a typo, there is some kind of underlying problem here worth investigating that
+  // makes switching the latitude and longitude a required step to produce the correct results.
+  const latitude = popUpInfo.properties.longitude ?? 0;
+  const longitude = popUpInfo.properties.latitude ?? 0;
```
SeanKennyNF, line 56: Highlighting one of those "questionable things" I mentioned at the start.
I don't know why this gives the correct results but it does.
I assume there's some kind of underlying problem here that needs investigation, I'd be happy to take this away and see why exactly it's like this and how we can clean it up. I'm hesitant about doing it here because I don't know how much of a rabbit hole it is so I was planning on leaving this as-is for this PR if that's alright with everyone.

components/ui/pathogen-map/pathogen-map-popup.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  const transformedProperties = Object.fromEntries(
+    Object.entries(popUpInfo.properties).map(([key, value]) => {
+      // Arrays in popUpInfo.properties that reach this point have been converted from
+      // ["ABC", "DEF"] to '["ABC", "DEF"]'
+      // In order to proceed further, they need to be converted back to arrays.
+      if (/^\[("[^\[\]\"]*",?)+\]$/.test(value)) {
```
SeanKennyNF, line 63: 🤮 

I'm like somewhat sure this something mapbox is doing that I don't have any control over and just have to undo. Again, this is one of those questionable things I'd be happy to investigate further, I just don't want to block on it right now if that's alright.

Obviously if you disagree, just say so but again I'm pretty sure this is some mapbox stuff at work and I'm not sure how much of a say I have about this. If you're curious, the `antibodies` field is the problematic one in this case.

If you'd like to make the case that needing this suggests that my types are wrong you're probably right so I might have to ask for a pass on that one too to take another look at that when I take a look at whether the whole issue above is actually something fixable or not

components/ui/pathogen-map/pathogen-map-layer.tsx
```diff
@@ -0,0 +1,46 @@
+import { Layer, Source } from "react-map-gl";
```
SeanKennyNF, line 1: The idea with these `pathogen-map` files is that the arbotracker map and the serotracker map can share the `<PathogenMap>` component and literally every other component here is just something that lives under `<PathogenMap>`. Let me know if you have any questions :+1:

components/ui/pathogen-map/pathogen-map-popup.tsx
```diff
@@ -0,0 +1,102 @@
...
+      return [key, value];
+    })
+  ) as VisiblePopupInfo<TPathogenDataPointProperties>["properties"];
+
+  return (
+    <Popup
+      // Important, without this key, clicking from one popup to another does not trigger the onOpen since react thinks the component data
+      // has changed instead of a new pop up being opened. If you choose to change this, please ensure that in your solution the map flys to
+      // the correct location when clicking on one data point while having the popup already open.
+      key={transformedProperties.id}
```
SeanKennyNF, line 82: If you're wondering why I didn't move the `map.flyTo` to `onMouseDown` if this is the case, for some reason it doesn't work if you put it in the `onMouseDown`. I tried for ages to make it work in the `onMouseDown` and I swear it's some kind of janky mapbox interaction because I'm convinced I was calling everything correctly. Honestly this looks cleaner anyways imo just leaving a comment in case you were curious.

components/ui/pathogen-map/pathogen-map.tsx
```diff
@@ -0,0 +1,102 @@
...
+  PathogenMapPopup,
+  PopupContentGenerator,
+  PopupInfo,
+} from "./pathogen-map-popup";
+import { PathogenMapLayer, PathogenMapLayerInfo } from "./pathogen-map-layer";
+
+export interface PathogenDataPointPropertiesBase {
+  id: string;
+  latitude: number | undefined;
+  longitude: number | undefined;
```
SeanKennyNF, line 28: Idea behind this type is that if you have data associated to a point on the map you need it to have _at least_ these properties. Keeping it to a minimum so that the arbo stuff and sero stuff can share this component. That's the same reason for all of the generic type stuff. I know you could technically remove the generic type stuff and make this work but I think that's because of how many `any`s are floating around in the code so I'd rather keep the generic around so that when those `any`s get assigned to the correct type my stuff will have very specific typing.

components/ui/pathogen-map/use-pathogen-map-mouse-events.ts
```diff
@@ -0,0 +1,70 @@
...
+export type PathogenMapCursor = "" | "pointer";
+
+interface UsePathogenMapMouseProps<TPathogenDataPointProperties extends PathogenDataPointPropertiesBase> {
+  baseCursor: PathogenMapCursor;
+  layers: PathogenMapLayerInfo<TPathogenDataPointProperties>[];
+  setPopUpInfo: (input: PopupInfo<TPathogenDataPointProperties>) => void
+}
+
+export const usePathogenMapMouse = <TPathogenDataPointProperties extends PathogenDataPointPropertiesBase>({ baseCursor, layers, setPopUpInfo }: UsePathogenMapMouseProps<TPathogenDataPointProperties>) => {
+  const [cursor, setCursor] = useState<string>(baseCursor);
```
SeanKennyNF, line 16: For the record, all this `cursor`, `setCursor`, `onMouseEnter`, `onMouseLeave` stuff is for is so that the cursor has the right icon when you hover over a point on the map.

Docs: https://visgl.github.io/react-map-gl/docs/api-reference/map#onmouseenter

contexts/arbo-context.tsx
```diff
@@ -61,44 +62,16 @@ function filterData(data: any[], filters: { [key: string]: string[] }): any[] {
...
+export const arboReducer = (state: ArboStateType, action: ArboAction, map: MapRef | undefined) => {
   switch (action.type) {
     case ArboActionType.UPDATE_FILTER:
       const selectedFilters = {
         ...state.selectedFilters,
         [action.payload.filter]: action.payload.value,
       };
 
-      if (action.payload.map) {
-        setMapboxFilters(selectedFilters, action.payload.map);
```
SeanKennyNF, line 99: I'm piping the filtered data down so doing anything additional with the mapbox filters is unecessary.

components/ui/pathogen-map/use-pathogen-map-mouse.ts
```diff
@@ -0,0 +1,68 @@
+import { useState } from "react";
```
SeanKennyNF, line 1: All this `cursor`, `setCursor`, `onMouseEnter`, `onMouseLeave` stuff is for is so that the cursor has the right icon when you hover over a point on the map.

Docs: https://visgl.github.io/react-map-gl/docs/api-reference/map#onmouseenter

app/pathogen/arbovirus/dashboard/(map)/MapAndFilters.tsx
```diff
@@ -1,14 +1,15 @@
 "use client";
```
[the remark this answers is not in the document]
HimanshuRanka: This is really awesome and interesting. I remember using a react mapbox tool in the past (but I do not think it was this) and deciding against it because it had a bunch of issues. Looking at the documentation for this is p impressive and having it all be React-like would be great

app/pathogen/arbovirus/dashboard/(map)/MapAndFilters.tsx
```diff
@@ -73,7 +68,40 @@ export default function MapAndFilters() {
             "w-full h-full overflow-hidden col-span-6 row-span-2 relative"
           }
         >
-          <CardContent ref={mapContainer} className={"w-full h-full p-0"} />
+          <div className={"w-full h-full p-0"}>
```
HimanshuRanka, line 71: W O W

components/ui/pathogen-map/pathogen-map-popup.tsx
HimanshuRanka: Sounds good to me, can add it in the issue from above. 

Note: antibodies has always been a problem child. Mapbox does have some issues with data points being arrays. The approach I think of here is converting the array to a string higher up and filtering on the string (changing filter logic from array.find() to string.hasSubstring() -- mehtods do not exist, only highlight logic LOL). We can work on this later though and this looks fine to me for now

components/ui/pathogen-map/pathogen-map-popup.tsx
```diff
@@ -0,0 +1,102 @@
...
+    let mapZoom = map.getZoom();
+
+    return 80 / Math.pow(2, mapZoom);
+  };
+
+  if (popUpInfo.visible === false) {
+    return null;
+  }
+
+  // Not a typo, there is some kind of underlying problem here worth investigating that
```
HimanshuRanka, line 53: Also an issue I have seen before (We had lat and long switched in the other map). I wonder if we are just calculating them incorrectly or smthn in the ETL

components/ui/pathogen-map/pathogen-map.tsx
```diff
@@ -0,0 +1,102 @@
...
+    );
+  }, []);
+
+  if (!mapStyle) {
+    return;
+  }
+
+  return (
+    <Map
+      id="arboMap"
```
HimanshuRanka, line 73: Note: Need to make this dynamic going forward if we want to use this for other pathogens as well

HimanshuRanka:
BUG: Go Arbovirus > Analysis > filters- producer: HAI -> Causes the application to fail. Seems like other filters are working fine though.

app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -150,22 +140,18 @@ export default function Filters(props: { map?: mapboxgl.Map | null }) {
               "Antibody",
               state,
               filters.data["antibody"],
-              map,
               data ? data.records : [],
             )}
           </div>
-          {!map && (
```
HimanshuRanka, line 157: We do not want the Pathogen filters to be visible when we are on the map side (since this is done in the legend). I believe we were checking if map was null or not before, but since that does not exist anymore the filters are always visible

HimanshuRanka:
I think another enhancement on the Tech side I would love to see here is if we could pull the <Filters /> component one level up so it does not rerender between the analysis and the dashboard pages anymore since we do not need to send it filters. This would automatically make MapAndFilters redundant since now we would have a decoupled Map and Filters. 

I would assume moving it to the layout one level up should work.

HimanshuRanka (changes requested):
LGTM overall. There is a small bug I found and a potential enhancement that would be great (Ok to create a seperate ticket for enhancement)  but the 2 filters issues should be solved here if poss since this PR is causing them. 

Lmk if for whatever reason this seems really challenging and we can consider adding them to a bug ticket instead

app/pathogen/arbovirus/dashboard/filters.tsx
SeanKennyNF: Makes sense, resolved in b140c6a09e7732a7122b7fa70465b57a78bd66c9 :)

cc @gurickkooner since I know you're doing some stuff in that file too. Taking a read of that commit might be worth your time and I'd be happy to resolve any merge conflicts between my stuff and your stuff.

app/pathogen/arbovirus/analyze/columns.tsx
```diff
@@ -5,6 +5,7 @@ import React from "react";
 import { Button } from "@/components/ui/button";
 import { ArrowUpDown } from "lucide-react";
 import { pathogenColors } from "../dashboard/(map)/MapAndFilters";
+import validator from "validator";
```
SeanKennyNF, line 8: Commenting on random line for thread @HimanshuRanka 

> BUG: Go Arbovirus > Analysis > Assay - producer: HAI -> Causes the application to fail. Seems like other filters are working fine though.
> Only exists on analysis side.

I checked this one out a bit further and the url is actually `-` in airtable. My change appears to have affected the random records to be shown in the table at the start and the record with `-` just happened to be in the first 10. If you go to the real pathotracker and specify "assay: HAI" and go forward like six pages or something you actually get the same error which is interesting.

Anyways, resolved in f0ce3822feb2e878cb6876354cf4391e44f20034 so we won't have to think about this anymore :)

SeanKennyNF, line 8: Commenting on random line for thread @HimanshuRanka 

> I think another enhancement on the Tech side I would love to see here is if we could pull the component one level up so it does not rerender between the analysis and the dashboard pages anymore since we do not need to send it filters. This would automatically make MapAndFilters redundant since now we would have a decoupled Map and Filters.
> I would assume moving it to the layout one level up should work.

Ouch actually NextJS makes this a bit more complicated than you might initially think it is. The real problem here is that I need it to be different between the pages and the layout doesn't help me at all here since it doesn't know what page it belongs. As far as decoupling Map and Filters even that doesn't come out as nice as you might initially think and I can't put it in the page because I need to pass props down to it now and NextJS doesn't like it if I do that. Maybe this is something we could discuss in person or maybe take a second look at later?

image

HimanshuRanka: Oh shit I had no idea about that LOL, guess that was a good catch for us then

HimanshuRanka (approved):

PR #58, opened 2023-12-04
Arbo vis

HimanshuRanka:
Added in several new visualizations using recharts as per this document: https://docs.google.com/document/d/1YsIj7Cv2cA2XY_QM4TKHeDFWsY3rBI-V/edit


next.config.js
```diff
@@ -1,4 +1,8 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {}
+const nextConfig = {
```
HimanshuRanka, line 2: This as well as the changes to the prefetch queries were all attempts to avoid the weird caching behaviour. This specifically prevents next from caching at all in a dev env. I have logged a ticket to investigate this further so we do not have random things for the same purpse everywhere in #62

app/pathogen/arbovirus/analyze/page.tsx
```diff
@@ -8,34 +8,53 @@ import {
...
-          <CountOfStudiesStratifiedByAntibodyAndPathogen />
+      <div className={"col-span-5 row-span-2 overflow-auto"}>
+        <VisualizationCard title={"Estimate count by WHO region and pathogen"}>
+          <WHORegionAndArbovirusBar />
+        </VisualizationCard>
+        <VisualizationCard title={"Estimate count by arbovirus & antibody type"}>
+          <AntibodyPathogenBar />
+        </VisualizationCard>
+        <VisualizationCard title={"Median seroprevalence by WHO Region"} height="h-full">
+          <MedianSeroPrevByWHOregion />
```
harrietware, line 42: For WHO region charts, can we remove estimates where WHO region is N/A? (Leave them on the map for now, but we may have to change this later for our WHO partnership) 

Also can we force the x-axis label to display all WHO regions?
Screen Shot 2023-12-04 at 9 59 00 AM

app/pathogen/arbovirus/analyze/recharts.tsx
```diff
@@ -0,0 +1,906 @@
...
+        <XAxis dataKey="year" />
+        <YAxis />
+        <Tooltip />
+        <Legend
+          layout="vertical"
+          verticalAlign="middle"
+          align="right"
+          wrapperStyle={{ right: -10 }}
+        />
+        <Area
```
SeanKennyNF, line 191: Similar kind of thing to my other comment on line 654 to line 782
We could have some kind of global object like this:
```
  const groupings = [
    {
      dataKey: "Zika",
      fillColor: pathogenColors.ZIKV,
    },
    {
      dataKey: "Dengue",
      fillColor: pathogenColors.DENV,
    },
    {
      dataKey: "Chikengunia",
      fillColor: pathogenColors.CHIKV,
    },
    {
      dataKey: "Yellow Fever",
      fillColor: pathogenColors.YF,
    },
    {
      dataKey: "West Nile",
      fillColor: pathogenColors.WNV,
    },
    {
      dataKey: "Mayaro",
      fillColor: pathogenColors.MAYV,
    },
  ] as const;
```
And make use of maps?
```
        {groupings.map(({ dataKey, fillColor }) => (
          <Area
            type="monotone"
            dataKey={dataKey}
            stackId="1"
            stroke={fillColor}
            fill={fillColor}
          />
        ))}
```

SeanKennyNF (approved):
Hey man, I've read through enough of this to have a solid grasp of what's going on and I do have a few more thoughts I want to leave here besides what I've already written. I'm just leaving my first few thoughts and I'll sit back down to write the rest in a few minutes here.

Full thoughts by this evening latest promise 🤞 <- pretend that's a pinkie promise emoji

app/pathogen/arbovirus/analyze/recharts.tsx
```diff
@@ -0,0 +1,906 @@
...
+//Study count by pathogen and antibody type
+
+type arbovirusesSF = "DENV" | "ZIKV" | "CHIKV" | "YF" | "WNV" | "MAYV";
+type arboviruses =
+  | "Dengue"
+  | "Zika"
+  | "Chikengunia"
+  | "Yellow Fever"
+  | "West Nile"
+  | "Mayaro";
```
SeanKennyNF, line 34: Might be worth coordinating with Elias + Gurick for a bit of consistency here. Ex. they have "Mayaro Virus" and "Zika Virus" whereas you haven't put virus here. Pretty small tbh just my consistency itch mostly hahaha

SeanKennyNF: Also I think "Chikungunya" might be the correct spelling but I'm open to being wrong here.

harrietware: +1

HimanshuRanka:
> Actually nah, this works. I agree with you that there's definitely enhancements in how we can write this and I was reading this and pretty sure there was a way for the visualizations to share a bunch of logic but I think you're right that the best avenue here is just to ship this and then make those enhancements as tech debt stuff 👍
> 
> Harriet's comments are super legit though and I think we should do those.
> 
> ~Hey man, I've read through enough of this to have a solid grasp of what's going on and I do have a few more thoughts I want to leave here besides what I've already written. I'm just leaving my first few thoughts and I'll sit back down to write the rest in a few minutes here.~
> 
> ~Full thoughts by this evening latest promise 🤞 <- pretend that's a pinkie promise emoji~

Yeah I totally agree with all your comments. Main reason a lot of it was hardcoded out was I am still figuring out the library and wanted to have full control on individual customization (the need for which dropped when we reached stable charts) but I just felt like refactoring that stuff rn would take a long time and I would want to be mindful of the whole charting ecosystem which would be a larger process I think. 

Will make the changes Harriet requested and can go ahead with it then.

SeanKennyNF:
> > Actually nah, this works. I agree with you that there's definitely enhancements in how we can write this and I was reading this and pretty sure there was a way for the visualizations to share a bunch of logic but I think you're right that the best avenue here is just to ship this and then make those enhancements as tech debt stuff 👍
> > Harriet's comments are super legit though and I think we should do those.
> > ~Hey man, I've read through enough of this to have a solid grasp of what's going on and I do have a few more thoughts I want to leave here besides what I've already written. I'm just leaving my first few thoughts and I'll sit back down to write the rest in a few minutes here.~
> > ~Full thoughts by this evening latest promise 🤞 <- pretend that's a pinkie promise emoji~
> 
> Yeah I totally agree with all your comments. Main reason a lot of it was hardcoded out was I am still figuring out the library and wanted to have full control on individual customization (the need for which dropped when we reached stable charts) but I just felt like refactoring that stuff rn would take a long time and I would want to be mindful of the whole charting ecosystem which would be a larger process I think.
> 
> Will make the changes Harriet requested and can go ahead with it then.

Sounds like a plan to me :+1:

harrietware (approved):
This is great! Last request - could we do the median seroprevalence graphs as a percentage on the y axes?

Issue #60, opened 2023-12-04
Add a feature flag that dictates whether or not the SARS-CoV-2 portions of the website are visible.


SeanKennyNF:
Based on the discussion here. https://github.com/PathoTracker/Pathotracker/pull/57#discussion_r1413269388

In anticipation for the conference next week, we would like to add a feature flag to hide the SARS-CoV-2 portions of the website so that only Arbotracker is visible that we can use to hide those portions in production but keep them on in our local environments.

PR #66, opened 2023-12-05
Added the `SARS_COV_2_TRACKER_ENABLED` feature flag

SeanKennyNF:
Resolves https://github.com/PathoTracker/Pathotracker/issues/60

Feature flag off:
FFFFFFFFF

Attempts to access `/pathogen/sarscov2/dashboard` and `pathogen/sarscov2/analyze` do not work and give 404s

HHHHHHHHHHHH

Feature flag on:
GGGGGGGG

Attempts to access `/pathogen/sarscov2/dashboard` and `pathogen/sarscov2/analyze` work


app/pathogen/sarscov2/layout.tsx
```diff
@@ -1,43 +1,48 @@
-import React from "react";
```
SeanKennyNF, line 1: Another gross diff :/

Only change is that I added
```

  if(!process.env.SARS_COV_2_TRACKER_ENABLED) {
    return notFound();
  }
```
on line 34. Everything else is the same.

SeanKennyNF, line 1: If you're wondering, when the flag isn't specified, `process.env.SARS_COV_2_TRACKER_ENABLED` is `undefined` which is falsy so this will have the right value even if we don't change anything in vercel hahaha

components/customs/header.tsx
```diff
@@ -105,6 +107,7 @@ export const Header = () => {
               </ul>
             </NavigationMenuContent>
           </NavigationMenuItem>
+          {/*
```
SeanKennyNF, line 110: Commented out "About" and "Documentation". Didn't feature flag off the "about" page because it felt unecessary but please let me know if that's something we'd like and I'll do it.

components/customs/header.tsx
```diff
@@ -119,15 +122,16 @@ export const Header = () => {
...
         </NavigationMenuList>
       </NavigationMenu>
 
-      {/*TODO: Need to incorporate language switching*/}
       <div
         className={
           "text-md font-header non-italic text-white h-100 flex justify-end items-center cursor-pointer"
         }
       >
+      {/* TODO: Need to incorporate language switching
```
SeanKennyNF, line 134: Don't want the language switcher to show up if we haven't implemented it lol

lib/country-bounding-boxes.ts
```diff
@@ -37,6 +37,7 @@ const countryBoundingBoxData: Record<string, CountryBoundingBox | undefined> = {
   Cameroon: [8.49, 1.73, 16.01, 12.86],
   "Democratic Republic of the Congo": [12.18, -13.26, 31.17, 5.26],
   "Republic of Congo": [11.09, -5.04, 18.45, 3.73],
+  "Congo": [11.09, -5.04, 18.45, 3.73],
```
SeanKennyNF, line 40: Just a quick band-aid because some of the countries were renamed in airtable so the country zoom was broken for those countries. Yes I know it's silly that I tied the name in airtable to the bounding box, I've got an idea that might (?) fix this eventually but you'll have to bear with this for now, sorry :bow:

EliasPinno (approved):
Before we merge this, @HimanshuRanka should set the feature flags status on prod. Though this will work on prod it is bad practice to have it unset.

HimanshuRanka:
> Before we merge this, @HimanshuRanka should set the feature flags status on prod. Though this will work on prod it is bad practice to have it unset.

Updated vercel with this set to false for prod and true for all other deployments

HimanshuRanka (approved):
LGTM

Issue #64, opened 2023-12-04
Convert the SARS-CoV-2 maps to also use the `PathogenMap` component.


SeanKennyNF:
> I know the CoV2 map still uses the raw mapbox-gl API. I tried to switch it to this one and I'm 100% sure it's possible and makes things a bunch cleaner but I don't want this PR to be massive so I stopped myself but I definitely intend to switch both maps over to this library so that we can delete basically the entire useMap file and all of the stuff that uses the mapbox-gl API directly.

From this comment https://github.com/PathoTracker/Pathotracker/pull/54.

We've switched over to using the `react-map-gl` library to make the map less buggy and we should switch over on the SARS-CoV-2 side of the app as well so that we can delete all the code that uses the old `useMap` hook which called the `mapbox-gl` API directly.

PR #71, opened 2023-12-06
Converted the SARS-CoV-2 map to use the `PathogenMap` component.

SeanKennyNF:
Resolves https://github.com/PathoTracker/Pathotracker/issues/64 and also resolves https://github.com/PathoTracker/Pathotracker/issues/35 for the SARS-CoV-2 map since this bug was also present there.

The SARSCoV2 stuff is kind of buggy but this shouldn't introduce any new problems at least hahaha. 

R.I.P. `useMap` 2023-2023

pr-71-demo


EliasPinno (changes requested):
I think we should deliberately not merge this before Heidelberg. Just no reason to do so if since I feel like it possible this could have side effects for arbotracker.

EliasPinno (approved):

SeanKennyNF:
Will merge once mapbox API key is working again.

PR #73, opened 2023-12-06
feat: sampling-date-filter

gurickkooner:
-Added start and end date filter fields that open up a calendar date picker 
-Data is then filtered based on selected start date, end date, and in between if both are selected (changes are reflected on the map)
-Added drop down to increase user experience allowing month and year to be selected
-Fixed  issues causing the the calendar date picker to jump around when navigating
-Added x icon to clear the start and end date fields
-Ensured that the reset button clears the date fields as well as the other fields

Below is a screenshot of the date filtering
Screenshot 2023-12-06 at 13-07-54 Pathotracker


SeanKennyNF:
> Should the target branch be changed to main since the issue-11-update-arbo-filter has been merged?

@gurickkooner yes :)

SeanKennyNF (approved):
Might be good to clean up some of the `console.log`s. Otherwise lgtm

PR #77, opened 2023-12-06
Made a bunch of small changes across the application

HimanshuRanka:
Made a lot of the simpler changes as requested in the duration of the call. There are a few more changes that will be done in quick succession that will be traced with tickets.


SeanKennyNF (approved):
Big thanks for cleaning up those `console.log`s too. I'm getting those tickets cooked up so please be patient :)

👨‍🍳

Issue #81, opened 2023-12-07
Sorting by "Sampling Start Date" or "Sampling End Date" on the table in the analysis tab incorrectly sorts the dates alphabetically instead of chronologically.


SeanKennyNF:
If you sort the sampling start date or sampling end date columns in the analysis tab, it incorrectly puts studies whose sampling end or start date is in April at the top since "A" is the first letter of the alphabet instead of putting studies whose start or end date is the furthest into the past at the top. We should fix this and those columns should sort by doing a date comparison instead of an alphabetical string comparison.

PR #91, opened 2023-12-08
Table date fixes

### LINKED ISSUE #78: Better handling for estimates without a sampling end date or sampling start date.
**Update: As of https://github.com/PathoTracker/Pathotracker/pull/77, the app no longer crashes when the modal is clicked on but every other issue listed here is still present.**

Every single estimate in Central African Republic has a start date but does not have an end date. Clicking on one of those estimates on the map in order to open up the modal crashes the application. We need to make sure that we handle estimates without a sampling start or end date better (I was unable to find estimates without a start date but we should still handle that just in case)

1. If a study doesn't have a sampling end date or sampling start date, the sampling date range displayed in the modal should just be the start date or the end date twice depending on which date is actually present (Ex. Mar 15, 2009 to Mar 15, 2009)
2. The filter should treat these studies as having a sampling range of one day instead of infinite days just to make sure that these studies aren't always present with the end date filter past a certain point just because we don't know when their sampling ended.
3. In the table, we could specify N/A for the sampling end date or sampling start date or we could just use the other value, that might be okay too.

Currently in the Analyze tab, the sampling end date listed for studies without a sampling end date is Dec 31, 1969, which is wrong and I'd take either the sampling start date or "N/A" or "-" or anything really over that.

Capture5555555555555555555

EliasPinno:
Fixed sorting to actually sort on date instead of alphabetically. 
Handled when dates are not given to instead return N/A.
Closes #78 and #81

### DIFF
--- a/app/pathogen/arbovirus/analyze/columns.tsx
+++ b/app/pathogen/arbovirus/analyze/columns.tsx
@@ -4,7 +4,7 @@ import { ColumnDef } from "@tanstack/table-core";
 import React from "react";
 import { Button } from "@/components/ui/button";
 import { ArrowUpDown } from "lucide-react";
-import { HeaderContext } from "@tanstack/react-table";
+import { HeaderContext, Row } from "@tanstack/react-table";
 import validator from "validator";
 import Link from "next/link";
 import { TranslateDate } from "@/utils/translate-util/translate-service";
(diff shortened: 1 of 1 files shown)


EliasPinno:
TableShowOff

SeanKennyNF (approved):

PR #93, opened 2023-12-08
Code review date range hot fix

gurickkooner:
Bugfix for the sampling date ranges to only compare using month and year not the specific date of a study to allow for more generic approach to filtering. 
datefilters

### DIFF
--- a/app/pathogen/arbovirus/dashboard/filters.tsx
+++ b/app/pathogen/arbovirus/dashboard/filters.tsx
@@ -51,7 +51,6 @@ const buildFilterDropdown = (
   state: ArboContextType,
   filterOptions: string[],
   data: any,
-  { key }: { key: string } // add key prop
 ) => {
   if (
     filter === FilterableField.start_date ||
@@ -141,7 +140,6 @@ const FilterSection = ({
           state,
           filters.data[field],
           data ? data.records : [],
-          { key: field } // add key prop
         );
       })}
     </div>


contexts/arbo-context.tsx
```diff
@@ -129,7 +127,9 @@ export const arboReducer = (
       if (map) {
         adjustMapPositionIfCountryFilterHasChanged(action, map);
       }
-
+      console.log("Data filters", action.payload.filter);
+      console.log("Data before filtration", state.filteredData);
+      console.log("Data after filtration", filterData(action.payload.data, selectedFilters));
```
SeanKennyNF, line 132: Let's maybe clean up the `console.log`s before we merge?

app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -50,7 +50,8 @@ const buildFilterDropdown = (
   placeholder: string,
   state: ArboContextType,
   filterOptions: string[],
-  data: any
+  data: any,
+  { key }: { key: string } // add key prop
```
SeanKennyNF, line 54: Is this prop valuable? I don't think it's used anywhere. The divs already have keys on them, can we just remove  this?

SeanKennyNF (commented):
Couple suggestions, looks like it works though.

app/pathogen/arbovirus/dashboard/filters.tsx
gurickkooner: I was getting a warning that there were no keys so i just put this in to get rid of that warning. I can take it out and see if what happens.

gurickkooner: Yeah appears to not to be valuable as you said the divs have the keys on them

SeanKennyNF (approved):

PR #94, opened 2023-12-09
Countries are shaded light blue if they contain any estimates and remain white if they do not.

### LINKED ISSUE #74: Country shading based on presence of estimates in that country
Countries with estimates present should be coloured light blue and countries without estimates present should be coloured white.
Refer to the map on the original SeroTracker https://serotracker.com/en/Explore to see what colours we should use.

SeanKennyNF:
Resolves #74

peek-08-12-2023-2


components/ui/pathogen-map/pathogen-map-layer.tsx
```diff
@@ -2,15 +2,40 @@ import { Layer, Source } from "react-map-gl";
 import { PathogenDataPointPropertiesBase } from "./pathogen-map";
 import { PathogenMapCursor } from "./use-pathogen-map-mouse";
 
-export interface PathogenMapLayerInfo<
+export interface PathogenMapLayerInfoWithoutCountryHighlighting<
   TPathogenDataPointProperties extends PathogenDataPointPropertiesBase
 > {
   id: string;
   cursor: PathogenMapCursor;
+  isDataUsedForCountryHighlighting: false;
```
SeanKennyNF, line 10: > Lets you have multiple layers but only one of them at a time is going to have their data used for country highlighting and the other ones don't need `country` to be a key on their data but the one you want to use does need that key in it's data :)

These types are that same idea.

lib/country-iso-3166-1-alpha-3-codes.ts
```diff
@@ -0,0 +1,195 @@
+export const countryNameToIso31661Alpha3CodeMap: Record<string, string> = {
```
SeanKennyNF, line 1: I think I'm probably going to start saving the three letter country codes in the backend at some point so that the names in Airtable aren't coupled to the react app but this will have to do for now.

Basically know that things better than this will be coming at some point so I wouldn't get too concerned about this one :)

components/ui/pathogen-map/pathogen-country-highlight-layer.tsx
```diff
@@ -0,0 +1,102 @@
...
+          !!alpha3CountryCode
+      ) ?? []
+  );
+
+  const countryColorsAndOpacities = Array.from(
+    allUniqueCountryCodesWithData
+  ).map((alpha3CountryCode) => ({
+    alpha3CountryCode: alpha3CountryCode,
+    fill: MapSymbology.CountryFeature.HasData.Color,
+    opacity: MapSymbology.CountryFeature.HasData.Opacity,
```
SeanKennyNF, line 46: Would be down to turn the opacity down a little bit or make the colour a little bit less harsh, right now it's 0.5 and you can see what that looks like in the gif (maybe a bit harsh with the pastel colours?)

EliasPinno (approved):

PR #95, opened 2023-12-11
Fixed a bug that was making filter on "Producer: Euroimmun" impossible.

### LINKED ISSUE #83: Filtering on "Producer: Euroimmun" is not possible.
Filtering by producers seems to work correctly but for some reason requests to filter on only records whose producer is "Euroimmun" are ignored. See gif (Euroimmun is being clicked on, not just hovered over):

peek33333333333333

SeanKennyNF:
Resolves #83

The software that makes gifs for me is broken right now so you'll have to go on Vercel yourself if you want to make sure my stuff works, really sorry about that one :sob:

image

### DIFF
--- a/components/customs/multi-select.tsx
+++ b/components/customs/multi-select.tsx
@@ -126,7 +126,10 @@ export function MultiSelect(props: MultiSelectProps) {
                       onSelect={(value) => {
                         setInputValue("");
                         const newValue = selectables.find(
-                          (option) => option.value === value,
+                          // CommandItem appears to strip starting and ending whitespace from the value given as a key
+                          // so the trim() fixes a situation where the value has some starting or trailing whitespace
+                          // and isn't recognized as a selectable option as a result.
+                          (option) => option.value.trim() === value,
                         );
                         if (newValue)
                           handleOnChange([...selected, newValue.label]);


components/customs/multi-select.tsx
```diff
@@ -126,7 +126,10 @@ export function MultiSelect(props: MultiSelectProps) {
                       onSelect={(value) => {
                         setInputValue("");
                         const newValue = selectables.find(
-                          (option) => option.value === value,
+                          // CommandItem appears to strip starting and ending whitespace from the value given as a key
+                          // so the trim() fixes a situation where the value has some starting or trailing whitespace
+                          // and isn't recognized as a selectable option as a result.
+                          (option) => option.value.trim() === value,
```
SeanKennyNF, line 132: For those who don't understand, the underling value for "Euroimmun" is `euroimmun ` and the underlying component library strips away whitespace when it sends it's `onSelect` event for some reason. All our logic still works with `euroimmun `, we just need to wrap this one particular event handler basically.

We could technically strip away the whitespace in the backend but it's kind of silly that you can't have any extra whitespace in your producer name imo so we should ship this anyways. I'm not sure if it's a super good use of our time to strip the whitespace on the backend but I could definitely make a ticket and we could hit it if there's interest.

HimanshuRanka (approved):
Oh damn! Thanks for getting this out of the way. 

And yeah I agree about whitespaces, no need to strip them in the backend. This solution ensures any additional data we add to the backend does not cause problems

PR #109, opened 2023-12-18
Issue 76 overlap filters

gurickkooner:
Have updated the filter logic so that it should work in all 4 cases. Which are depicted in the pdf shown below. 
One thing to note is that first the code for start_date filtering will be run and then once an end date is selected it runs the code for end_date. So this means that the data in the end_date if statement if already pre-filtered using the start date logic.
I have also added an if statement check for the awkward case where the end date was defaulting to 1969 so in the case where item end date is before item start date for any study it will set it to be 1 month after to allow for a period of overlap. Also attached some screenshots of overlapping periods.  

[Pathotracker.pdf](https://github.com/PathoTracker/Pathotracker/files/13851606/Pathotracker.pdf)
Screenshot from 2024-01-06 15-09-51
Screenshot from 2024-01-06 14-11-02
overlap1


contexts/arbo-context.tsx
```diff
@@ -47,51 +47,35 @@ function filterData(data: any[], filters: { [key: string]: string[] }): any[] {
...
-        );
-
-        const itemDateUTC = Date.UTC(
-          itemDate.getUTCFullYear(),
-          itemDate.getUTCMonth(),
-        );
-        
-        return itemDateUTC <= filterEndUTC;
+        // Check for any overlap in the sampling period
+        return !(itemEndDate < filterEndDate && itemStartDate < filterEndDate);
```
SeanKennyNF, line 61: Are we sure about this logic?

`itemEndDate` is always after `itemStartDate` so if `filterEndDate` is bigger than `itemEndDate` it's automatically bigger than `itemStartDate` so we can simplify this to this:

`!(itemEndDate < filterEndDate && itemStartDate < filterEndDate)` -> `!(itemEndDate < filterEndDate)`

Which is the same as this: `itemEndDate > filterEndDate`
The condition we had before was `filterEndDate <= filterEndUTC`
So it looks to me like we've just inverted our logic here? I'm assuming the start date is the same thing.

I checked out your branch and tried using out the filter and it looks like I'm right that there's something wrong with the logic here. I gave a sampling start date of 1982 and a sampling end date of 2014 which should return a whole bunch of studies but it returned zero because I think it's looking for everything that started before 1982 and ended after 2014 instead of everything which started at some point between 1982 and 2014

image

contexts/arbo-context.tsx
```diff
@@ -11,7 +11,7 @@ import {
   getBoundingBoxFromCountryName,
 } from "@/lib/country-bounding-boxes";
 import useArboData from "@/hooks/useArboData";
-import { parseISO } from "date-fns";
+import { parseISO, isAfter, isBefore } from "date-fns";
```
SeanKennyNF, line 14: :heart: :heart: :heart:
I love using these functions over `<` and `>` as well but it looks like you might have imported them and forgot to use them?

contexts/arbo-context.tsx
```diff
@@ -47,51 +47,35 @@ function filterData(data: any[], filters: { [key: string]: string[] }): any[] {
     return filterKeys.every((key: string) => {
       if (!filters[key].length) return true;
 
-      if(key === "end_date") {
-        const filterEndDate = parseISO(filters["end_date"][0]);
+      if (key === "end_date") {
+        const filterEndDate = new Date(filters["end_date"][0]);
```
SeanKennyNF, line 51: I'd like to keep the `parseISO` here so that we don't run into locale related bugs again. If there's a particular reason we need to use `new Date` here, let's fix that underlying issue and store iso strings instead.

SeanKennyNF (commented):
I don't think the logic here is correct.

contexts/arbo-context.tsx
gurickkooner: Will double check I had it traced out on paper but will double check. Yeah that is odd maybe it was doing the reverse. Weird i thought it was working for some of the date ranges I tried. This condition to get all the overlaps is kinda confusing haha. Will attempt to re-format it to get all the 4 conditions

contexts/arbo-context.tsx
gurickkooner: Oh yeah I tried to use those and for some reason it was giving me invalid dates I think it might be a specific formatting issue. I read the docs and was not entirely sure what they wanted.

contexts/arbo-context.tsx
gurickkooner: Fro sure will attempt to use parseISO i just kept running into issues with it but maybe that was because I was not consistent everywhere with passing the dates.

contexts/arbo-context.tsx
```diff
@@ -47,51 +47,44 @@ function filterData(data: any[], filters: { [key: string]: string[] }): any[] {
...
+        const itemEndDate = new Date(item.sample_end_date);
 
-        const filterEndUTC = Date.UTC(
-          filterEndDate.getUTCFullYear(),
-          filterEndDate.getUTCMonth(),
+        // Check for any overlap in the sampling period
+        return (
+          itemEndDate <= filterEndDate ||
+          (itemEndDate >= filterEndDate && itemStartDate < filterEndDate)
         );
```
SeanKennyNF, line 65: I'm curious if this is equivalent to `itemStartDate < filterEndDate`. I really like what you did with the other one with `itemEndDate >= filterStartDate` and I'm pretty sure that would work here if you also threw in your solution for the studies in the Central African Republic.

contexts/arbo-context.tsx
```diff
@@ -47,51 +47,44 @@ function filterData(data: any[], filters: { [key: string]: string[] }): any[] {
...
-        const itemDateUTC = Date.UTC(
-          itemDate.getUTCFullYear(),
-          itemDate.getUTCMonth(),
-        );
-        
+        // Check if the end date is before the start date (Fix for particular yellow fever studies in central africa that have start date 2009 and end date for 1969)
+        if (itemEndDate < itemStartDate) {
+          // Set the end date to be the same or 1 month after the start date
+          itemEndDate = new Date(itemStartDate);
+          itemEndDate.setMonth(itemEndDate.getMonth() + 1);
```
SeanKennyNF, line 82: My concern here would be instances where the item end date is in December. I would recommend something like the `addMonths` function from `date-fns` instead [docs](https://date-fns.org/v3.2.0/docs/addMonths)

SeanKennyNF (approved):
I tried playing around with this and it looks like it works to me :)

I left two suggestions I would probably consider. One suggestion is to fix a potential future bug and one suggestion is to simplify the code a bit.

PR #113, opened 2023-12-29
Switch the Arbovirus Portion of the website to using iit-backend v2

SeanKennyNF:
Please check out the message I dropped in basecamp if you need some context here!
https://3.basecamp.com/5061858/buckets/30589341/messages/6890211607

The functionality of the front end should be completely unchanged before and after this PR (with one exception which I've outlined) so this should be pretty seamless but let me know if you see any difference and we'll see what we can do!

I've played around with a local deployment fair amount so I'm pretty confident so feel free to run this on your own local machine if you'd like to build some confidence in it too!

### NOTE
**We need to set `NEXT_PUBLIC_API_GRAPHQL_URL` in the vercel environment variables to `https://iit-backend-v2.vercel.app/api/graphql` before the vercel deploy will work so if someone with vercel access could do that for me that would be great :heart: :heart: :heart:**

Peek 2023-12-31 20-07

Peek 2023-12-31 20-09


app/pathogen/arbovirus/layout.tsx
```diff
@@ -11,26 +16,14 @@ export default async function ArboLayout({
   const queryClient = getQueryClient();
 
   await queryClient.prefetchQuery({
-    queryKey: ["ArbovirusRecords"],
-    queryFn: () =>
-      fetch(`${process.env.NEXT_PUBLIC_API_BASE_URL}/arbo/records`, { cache: 'no-store' }).then(
-        (response) => response.json(),
-      ),
+    queryKey: ["arbovirusEstimatesQuery"],
+    queryFn: () => request(process.env.NEXT_PUBLIC_API_GRAPHQL_URL ?? '', arbovirusEstimatesQuery)
```
SeanKennyNF, line 20: @HimanshuRanka, if you're okay with the name could you please set `NEXT_PUBLIC_API_GRAPHQL_URL` to `https://iit-backend-v2.vercel.app/api/graphql` in all environments for me please? The vercel deployment won't work until you do hahahaha

app/pathogen/arbovirus/analyze/columns.tsx
```diff
@@ -238,43 +238,43 @@ export const columns: ColumnDef<Estimate>[] = [
...
+      const end_date = row.original.sampleEndDate;
       if(end_date) {
         return TranslateDate(end_date)
       }
       return 'N/A';
     }
   },
   {
-    accessorKey: "who_region",
+    accessorKey: "whoRegion",
```
SeanKennyNF, line 249: Alright so there's actually one tiny functionality difference here. The names of the columns in the dropdown and the csv downloaded are in camel case now instead of snake case. These names should really be "Sample Size" instead of "sample_size" or "sampleSize". I didn't make the change to keep this PR small so if everyone else is on board to hit this later, let's hit it later. Otherwise, just ask and I'll open a different PR that we can merge before this one that changes the names of the columns to be more friendly. I'm pretty flexible hahaha

image

app/pathogen/arbovirus/analyze/nivo-vis.tsx
SeanKennyNF, line 1: Deleted this file because it was making an API call and was unused.

EliasPinno (approved):

app/pathogen/arbovirus/analyze/columns.tsx
SeanKennyNF: Will do! :sparkles:

PR #114, opened 2024-01-02
Moved the download data button.

EliasPinno:
image
Image self explanatory.


components/ui/data-table.tsx
```diff
@@ -102,6 +102,9 @@ export function DataTable<TData, TValue>({
     <div>
       <div className="flex items-center py-4">
         <h2><b>Explore arbovirus seroprevalence estimates in our database</b></h2>
+        <Button variant="outline" className="ml-auto bg-foreground" onClick={getAllVisibleData}>
+          Download CSV
+        </Button>
```
EliasPinno, line 107: Because it is in a new spot, I get to reduce the required code too!

SeanKennyNF (approved):
😎

PR #125, opened 2024-01-09
Added the ability to zoom in on a visualization.

### LINKED ISSUE #118: Clicking on a visualizations expands it
When u click on a visualization it expands to take up the full screen size with more details.

SeanKennyNF:
Resolves #118. Adds the ability to zoom in on a visualization by clicking on it.

### DIFF
--- a/app/pathogen/arbovirus/analyze/page.tsx
+++ b/app/pathogen/arbovirus/analyze/page.tsx
@@ -8,18 +8,23 @@ import { AntibodyPathogenBar, MedianSeroPrevByWHOregion, MedianSeroPrevByWHOregi
 import clsx from "clsx";
 import { CardConfiguration, CardStyle, CardType, getConfigurationForCard } from "@/components/ui/card-collection/card-collection-types";
 import { CardCollection } from "@/components/ui/card-collection/card-collection";
+import { VisualizationId, addToVisualizationInformation } from "../visualizations/visualizations";
+import { ZoomIn } from "lucide-react";
+import { useRouter } from "next/navigation";
 
 const VisualizationCard = (props: {
   title: string;
   children: React.ReactNode;
   height?: string;
+  renderIcons: () => React.ReactNode;
 }) => {
-  
-  
   return (
     <Card className={clsx("mb-4 mr-4 p-4 pt-0", props.height ?? "h-full sm:h-3/4 2xl:h-1/2 relative")}>
       <CardContent className={"px-0 h-full flex flex-col"}>
-        <h3 className="py-4 w-full text-center text-lg">{props.title}</h3>
+        <div className="flex py-4">
+          <h3 className="w-full text-center text-lg">{props.title}</h3>
+          {props.renderIcons()}
+        </div>
         {props.children}
       </CardContent>
     </Card>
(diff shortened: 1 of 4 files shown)


utils/referrer-link-util.ts
```diff
@@ -0,0 +1,5 @@
+const safeReferrerLinkRegex = /^(\/[a-zA-Z0-9]+)+$/
+
+export const isSafeReferrerLink = (referrerLink: string) => {
+  return safeReferrerLinkRegex.test(referrerLink);
+}
```
SeanKennyNF, line 5: This is to prevent people from sending other people links like `https://serotracker-git-issue-118-clicki-abe9e2-himanshurankas-projects.vercel.app/pathogen/arbovirus/visualizations?visualization=median-seroprevalence-by-who-region&referrerRoute=https://airtable.com/` Where `https://airtable.com/` is a link to some kind of malicious website that you want them to visit when they click back. All I'm doing here is guaranteeing that we're going to somewhere internal.

lib/utils.ts
```diff
@@ -4,3 +4,7 @@ import { twMerge } from "tailwind-merge"
 export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs))
 }
+
+export const typedObjectEntries = <TKey extends string, TValue>(input: Record<TKey, TValue>): [TKey, TValue][] => {
```
SeanKennyNF, line 8: Default return type is `[string, TValue]` which isn't ideal for me. `[TKey, TValue]` is more strict and accurate.

app/pathogen/arbovirus/analyze/page.tsx
```diff
@@ -4,61 +4,77 @@ import React, { useCallback, useMemo } from "react";
...
+          <VisualizationCard 
+            key={visualizationInformation.id}
+            title={visualizationInformation.displayName}
+            height={visualizationInformation.height}
+            renderIcons={() => 
+              <button
+                onClick={() => router.push(`visualizations?visualization=${visualizationInformation.urlParameter}&referrerRoute=/pathogen/arbovirus/analyze`)}
+                aria-label="See visualization in fullscreen"
+              >
+                <ZoomIn />
```
SeanKennyNF, line 59: I think the zoom button is really cute because
1. We can make more clickable things on the visualization if we'd like
2. You actually know what's happening if you click on the button as opposed to clicking on the visualization.
3. I think it would be kind of cute to throw a download button or something next to it potentially for when we want to do the images
4. Getting a `onClick` for the entire visualization seems like a pain

app/pathogen/arbovirus/visualizations/page.tsx
```diff
@@ -0,0 +1,49 @@
+"use client"
```
SeanKennyNF, line 1: I love the idea of this as a separate page so that
1. Our page doesn't get huge and laggy (I also think some kind of solution with a modal would be more of a headache than this)
2. Someone can link a zoomed in version of a graph to someone else (Also helpful for citations?)
3. We can throw some more stuff on the page which I think would be fun!

Honestly Wikipedia when you want to click on an image to view it fullscreen was my real inspo here and I really liked how it turned out! It doesn't add load time, you can check out the vercel deploy.

EliasPinno (approved):
We have weird zoom in issues with the entire application, that are isolated from this PR. Using a scrollbar to handle them works fine for now before we move to prod. Will log as separate ticket.

app/pathogen/arbovirus/analyze/page.tsx
HimanshuRanka: yEAH i REALLY LIKE THE zoom button, it is more intuitive for sure!

app/pathogen/arbovirus/visualizations/page.tsx
HimanshuRanka: Oooo I really like this especially for our analytics even to see which graphs are being used

app/pathogen/arbovirus/visualizations/visualizations.tsx
```diff
@@ -0,0 +1,170 @@
...
+    urlParameter:
+      VisualizationUrlParameter[
+        "cumulative-estimate-count-over-time-by-arbovirus"
+      ],
+    displayName:
+      VisualizationDisplayName[
+        "Cumulative estimate count over time by arbovirus"
+      ],
+    renderVisualization: StudyCountOverTime,
+    classNameWhenFullscreen: "p-16"
```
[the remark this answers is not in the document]
HimanshuRanka: Yeah I am really starting to dislike recharts LMFAO

PR #126, opened 2024-01-11
Added a filter for UN Region

### LINKED ISSUE #116: Add filtering based on UN subregions
We want to add a filter that will allow users to filter on WHO subregions as well. Similar to WHO region when a subregion is selected we want the map to zoom in on that subregion. 

We want to make the subregion filter responsive to changes in the WHO region and country filters and vice versa.
- i.e. If we select a specific country, then the other 2 dropdowns only show options in which that country belong etc. Ask me for more details if confused. 

This should be additive. i.e. if multiple regions are selected the map zoom should include all the selected regions. 

https://www.google.com/url?q=https://unstats.un.org/unsd/methodology/m49/&sa=D&source=docs&ust=1704240536012873&usg=AOvVaw0kjgxUQNJsRAAgnRGNTTdG

SeanKennyNF:
Resolves #116. I've added a filter for UN region.

- There's a requirement in the ticket to also make selecting a UN or WHO region change the list of WHO regions or UN regions available. I'm going to gloss over that detail for a sec and ask for clarifications in the issue.
- There's a requirement for map zoom as well, I'm going to hit that in separate PR just to keep things clean.

Related PRs in release order
- https://github.com/SeanKennyNF/iit-backend-v2/pull/4
- https://github.com/PathoTracker/Pathotracker/pull/126
- https://github.com/PathoTracker/Pathotracker/pull/127

Peek 2024-01-12 12-41


components/customs/multi-select.tsx
```diff
@@ -15,22 +15,23 @@ export type MultiSelectOption = {
...
   selected: string[];
   handleOnChange: (selected: string[]) => void;
 }
 
-const createMultiSelectOptionList = (options: string[]) => {
+const createMultiSelectOptionList = (options: string[], optionToLabelMap: Record<string, string | undefined>) => {
   return options.map((option: string) => {
     return {
-      label: option,
-      value: option.toLowerCase(),
```
SeanKennyNF, line 26: Removal of `toLowerCase` was to move it to a place that was more appropriate and made more sense given what this line is actually accomplishing (it's similar to the `trim` so I moved it next to it)

components/customs/multi-select.tsx
```diff
@@ -128,11 +129,11 @@ export function MultiSelect(props: MultiSelectProps) {
                         const newValue = selectables.find(
                           // CommandItem appears to strip starting and ending whitespace from the value given as a key
                           // so the trim() fixes a situation where the value has some starting or trailing whitespace
-                          // and isn't recognized as a selectable option as a result.
-                          (option) => option.value.trim() === value,
+                          // and isn't recognized as a selectable option as a result. This is the same for the toLowerCase().
+                          (option) => option.label.trim().toLowerCase() === value.trim().toLowerCase(),
```
SeanKennyNF, line 133: Kind of interesting but this component appears to be using the label provided, not the value so I'm pulling the label value that the component is giving me and trying to find the value associated.

EliasPinno (approved):
Approved!

PR #127, opened 2024-01-12
Made the map zoom work with the UN region filter

### LINKED ISSUE #116: Add filtering based on UN subregions

SeanKennyNF:
Resolves #116. This makes the map zoom focus on the UN region when you pick one. I left a discussion in #116 about how the filtering with country and UN region doesn't play super well together but I don't think that's appropriate to hit as a part of this ticket since that's an issue already present with countries and WHO regions.

Related PRs in release order
- https://github.com/SeanKennyNF/iit-backend-v2/pull/4
- https://github.com/PathoTracker/Pathotracker/pull/126
- https://github.com/PathoTracker/Pathotracker/pull/127

Peek 2024-01-12 12-43


lib/bounding-boxes.ts
```diff
@@ -196,31 +222,214 @@ const countryBoundingBoxData: Record<string, CountryBoundingBox | undefined> = {
   "Saint Lucia": [-61.07, 13.72, -60.85, 14.1],
 };
 
+const alphaTwoCountryCodeBoundingBoxData: Record<
```
SeanKennyNF, line 225: Made this because I don't want country names changing to break anything anymore :sob:

Ideally, we switch the country stuff to use this one as well and then delete the country one entirely. I'd love to do that and that was kind of my secret motive for adding `alphaTwoCountryCode` on the API but one thing at a time.

lib/bounding-boxes.ts
```diff
@@ -196,31 +222,214 @@ const countryBoundingBoxData: Record<string, CountryBoundingBox | undefined> = {
...
+  typedObjectFromEntries(
+    Object.values(UNRegion).map((unRegion) => [
+      unRegion,
+      combineBoundingBoxes(
+        getCountryCodesInUNRegion({ unRegion })
+          .map((countryCode) => alphaTwoCountryCodeBoundingBoxData[countryCode])
+          .filter(
+            <T>(boundingBox: T | undefined): boundingBox is T => !!boundingBox
+          )
+      ),
```
SeanKennyNF, line 429: :heart: :heart: :heart:

lib/un-regions.ts
```diff
@@ -21,4 +21,263 @@ export enum UNRegion {
   MELANESIA = "MELANESIA",
   MICRONESIA = "MICRONESIA",
   POLYNESIA = "POLYNESIA",
-}
\ No newline at end of file
+}
+
+const countryAlphaTwoCodeToUNRegionsMap: Partial<Record<string, UNRegion>> = {
```
SeanKennyNF, line 26: Little bit of copy-pastrami from the backend

utils/typed-library-functions.ts
```diff
@@ -0,0 +1,5 @@
+export const typedObjectFromEntries = <TKey extends string, TValue>(
+  input: [TKey, TValue][]
+): Record<TKey, TValue> => {
+  return Object.fromEntries(input) as Record<TKey, TValue>;
+};
```
SeanKennyNF, line 5: I cry every time :smiling_face_with_tear:

contexts/arbo-context.tsx
```diff
@@ -157,27 +156,65 @@ export const arboReducer = (
...
 };
 
-const adjustMapPositionIfCountryFilterHasChanged = (
-  action: ArboAction,
-  map: MapRef
+interface GetAllBoundingBoxesFromExistingFiltersInput {
+  selectedFilters: Record<string, string[] | undefined>;
+}
+
+const getAllBoundingBoxesFromExistingFilters = (
```
SeanKennyNF, line 163: This part might be kind of hard to follow but basically the old code was pretty hacky and made an assumption that `country` was the only filter that could move the map. I need to look at the added filter as well as any existing filters now because there are two filters that determine the bounding box the map uses now. This isn't perfect looking either but yeah.

Q: Sean isn't this silly? I mean if you pick region "Eastern Europe" and "India" you'll filter out literally everything. Who cares if the map zooms in on the correct bounding box?

A: Yes you're right! This I left a discussion point on #116. I'd like to make a separate ticket to make any filters on geographical regions additive with one another so that "Eastern Europe" + "India" doesn't mean get me any "Indian  studies in Eastern Europe" and instead means "get me any studies from India or Eastern Europe".

EliasPinno (approved):

PR #138, opened 2024-01-20
Added the "Change in median seroprevalence" graph

### LINKED ISSUE #129: Create a visualization for seeing change in median seroprevalence over time
We should create a new visualization that shows change in median seroprevalence over time. Here's an image showing what that might look like. It might be helpful to have a ui component that lets you select how large the bins are (5 years, 1 year, 10 years?). This visualization would get affected by filters just like all of the other ones and just like "Median seroprevalence of arboviruses by WHO region", if you filter out specific arboviruses, their graph would no longer appear and the other graphs would grow in size to accomodate the new space.

image

This visualization should live in the analytics tab.

SeanKennyNF:
Resolves #129.
Pending feedback [here](https://3.basecamp.com/5061858/buckets/30589330/messages/6958342546) but would love to hear from the lovely dev team too :heart: :heart: :heart:

Peek 2024-01-20 15-32


lib/date-utils.ts
```diff
@@ -0,0 +1,13 @@
...
+export type TimeInterval = { intervalStartDate: Date; intervalEndDate: Date };
+
+export const doTimeIntervalsOverlap = (
+  intervalA: TimeInterval,
+  intervalB: TimeInterval
+): boolean => {
+  return (
+    isBefore(intervalA.intervalStartDate, intervalB.intervalEndDate) &&
+    isBefore(intervalB.intervalStartDate, intervalA.intervalEndDate)
+  );
```
SeanKennyNF, line 12: Proof this logic works in image below just because I know we've had a lot of conversations about what exactly the right way to do this is. I'll see if I can switch over the sampling date filter to use this piece of code as well at a later date :)

lib/utils.ts
```diff
@@ -1,10 +1,19 @@
...
   return Object.entries(input) as [TKey, TValue][];
+}
+
+export const typedObjectFromEntries = <TKey extends string, TValue>(input: [TKey, TValue][]): Record<TKey, TValue> => {
+  return Object.fromEntries(input) as Record<TKey, TValue>;
+}
+
+export const typedGroupBy = <TKey extends string, TValue>(values: TValue[], groupingFunction: (value: TValue) => TKey): Record<TKey, TValue[]> => {
+  return groupBy(values, groupingFunction) as Record<TKey, TValue[]>
 }
```
SeanKennyNF, line 19: My types are more correct than the library ones in these instances even though it needs a cast. This will matter once we have a non-`any` type on `filteredData` which I'd like to do using a graphql typegen library at a later date :)

app/pathogen/arbovirus/analyze/recharts/change-in-median-seroprevalence-over-time-graph.tsx
```diff
@@ -0,0 +1,196 @@
...
+              ? "h-1/2"
+              : "h-1/3";
+
+          return (
+            <div
+              className={clsx(width, height)}
+              key={`change-in-med-sero-prev-${arbovirus}`}
+            >
+              <h2 className="w-full text-center ">
+                {convertArboSFtoArbo(arbovirus as arbovirusesSF)}
```
SeanKennyNF, line 159: Sorry, I'm hoping to use a graphql typegen library on this repo eventually that should hopefully let us get better types and not have to do this.

app/pathogen/arbovirus/analyze/recharts/change-in-median-seroprevalence-over-time-graph.tsx
```diff
@@ -0,0 +1,196 @@
...
+          },
+        }).groupedDataPoints,
+      ]
+    )
+  );
+
+  return (
+    <div className="h-full flex flex-col">
+      <div className="h-full flex flex-row flex-wrap">
+        {Object.keys(dataGroupedByArbovirusThenByTimeBucket).map((arbovirus, index) => {
```
SeanKennyNF, line 136: Fair amount of copy-pistachio from "Median seroprevalence by WHO region and age group" if you recognize some of this.

app/pathogen/arbovirus/analyze/recharts/change-in-median-seroprevalence-over-time-graph.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+const generateBucketsForTimeInterval = (input: GenerateBucketsForTimeIntervalInput): GenerateBucketsForTimeIntervalOutput => {
+  let currentIntervalStartDate = startOfYear(input.earliestDate);
+  const buckets: TimeInterval[] = [];
+  
+  while(isBefore(currentIntervalStartDate, input.latestDate)) {
+    const currentIntervalEndDate = endOfYear(addYears(currentIntervalStartDate, input.groupingBucketSize.years));
+    const bucketToAdd = {
+      intervalStartDate: parseISO(currentIntervalStartDate.toISOString()),
+      intervalEndDate: parseISO(currentIntervalEndDate.toISOString())
```
SeanKennyNF, line 57: Scared of some JavaScript "copy-reference-instead-of-value"  thing so this is just my way of cloning the dates.

EliasPinno (approved):

PR #139, opened 2024-01-21
Issue 87 filter on who region

gurickkooner:
Upon selecting a who region map will pan over to that region. It will pan over to the region that  has been most recently selected ex: AFR, SEAR ( will  result in the map being panned over to SEAR)

### DIFF
--- a/contexts/arbo-context.tsx
+++ b/contexts/arbo-context.tsx
@@ -1,10 +1,6 @@
 "use client";
 
-import React, {
-  createContext,
-  useReducer,
-  useEffect
-} from "react";
+import React, { createContext, useReducer, useEffect } from "react";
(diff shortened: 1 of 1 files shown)


app/pathogen/arbovirus/dashboard/(map)/MapAndFilters.tsx
SeanKennyNF, line 1: Same thing, let's remove this empty file.

SeanKennyNF (commented):
Discussed in person, let's make the behaviour consistent with the UN region and country stuff where it sums up all of them and tries to make a box that fits all of them. Example is here https://github.com/PathoTracker/Pathotracker/pull/127, I'd be happy to answer any questions about the approach!

Issue #136, opened 2024-01-16
Consolidate the Dashboard and the Analyz pages


HimanshuRanka:
Use the Figma here to consolidate the dashboard and analyze page

https://www.figma.com/file/BKtEVRMNkZYh4EbG9T5wp5/PathoTracker?type=design&node-id=309-74&mode=design&t=pahP1zEB6Sp9J4Th-4

The final result should look like the frames below the large one

we want there to scroll snapping. So if a user were to begin scrolling, if they scrolled past the 33% point then the site would automatically snap to the next frame. Similar when scrolling up. 

Additionally we want the 3 sections to be navigatable to. Using an html ID param so that we can navigate to a specific part using the nav bar.

HimanshuRanka:
It will be important to consider the impact this will have on our state management system. Would we even continue to need it? Is there any use case where the user would navigate away from a dashboard and back expecting the filters to remain? Would we want to consider using local storage to save the filter values when a user navigates away from a page, and then load them back in when navigated to the page? 

I think this would be open to discussion so a part of this ticket would be to figure out the implications of the above consolidation and presenting them to the team. Naturally if the resolution seems straightforward then feel free to go ahead and implement it

HimanshuRanka:
Another thing to keep in mind is that the data section and the visualizations sections also look slightly different. Those changes are also included in here

HimanshuRanka:
This is a big boy ticket

SeanKennyNF:
Open to discussion but 
- I think it's a good idea to keep our context + state management stuff, I really don't want to prop drill a bunch of stuff and I like the setup we have. There's also the visualization pages and potentially more in the future that need that state so let's leave that as-is.
- Let's make a separate ticket if we want the filters to persist in local storage or a cookie or something, to me that's not mutually exclusive with keeping the state management system and doesn't have to be tied in with the redesign.

PR #141, opened 2024-01-21
Added skeleton of the new consolidated dashboard + analyze page with scroll snapping.

SeanKennyNF:
This PR is the first of many that will implement #136.
Here I'm just creating the skeleton of the page I'm going to build on. The really big feature here is the scroll snapping which I have put a detailed explanation of in my own review.

This page is feature flagged off obviously so nobody will be seeing this in production :)

### DIFF
--- a/app/pathogen/arbovirus/wip/page.tsx
+++ b/app/pathogen/arbovirus/wip/page.tsx
@@ -7,7 +7,7 @@ import { redesignedArbovirusPageSections, sectionUrlParameterToSectionId } from 
 export default function RedesignedArbovirusPage() {
   const searchParams = useSearchParams();
 
-  const {renderScrollSectionGroup, moveScrollSectionGroupToSection} = useScrollSectionGroup({
+  const { renderScrollSectionGroup, moveScrollSectionGroupToSection } = useScrollSectionGroup({
     scrollSectionGroupProps: {
       sections: redesignedArbovirusPageSections,
       className:"col-span-10 row-span-2",
@@ -17,7 +17,7 @@ export default function RedesignedArbovirusPage() {
 
   useEffect(() => {
     const sectionUrlParameter = searchParams.get('section');
-    const sectionId = sectionUrlParameterToSectionId({urlParam: sectionUrlParameter})
+    const sectionId = sectionUrlParameterToSectionId({ urlParam: sectionUrlParameter })
 
     if(sectionId) {
       moveScrollSectionGroupToSection({ sectionId });
--- a/app/pathogen/arbovirus/wip/page.tsx
+++ b/app/pathogen/arbovirus/wip/page.tsx
@@ -1,6 +1,6 @@
 "use client"
 import { useScrollSectionGroup } from "@/components/ui/scroll-section-group/use-scroll-section-group";
-import { useSearchParams } from "next/navigation";
+import { notFound, useSearchParams } from "next/navigation";
 import React, { useEffect } from "react";
 import { redesignedArbovirusPageSections, sectionUrlParameterToSectionId } from "./sections";
 
@@ -23,6 +23,10 @@ export default function RedesignedArbovirusPage() {
       moveScrollSectionGroupToSection({ sectionId });
     }
   }, [searchParams])
+
+  if(process.env.NEXT_PUBLIC_WEBSITE_REDESIGN_ENABLED !== 'true') {
+    return notFound();
+  }
   
   return (
     <div className="col-span-12 row-span-2 grid gap-0 grid-cols-12 grid-rows-2 grid-flow-col w-screen overflow-hidden border-box -my-4 -ml-4">
--- a/app/pathogen/arbovirus/wip/page.tsx
+++ b/app/pathogen/arbovirus/wip/page.tsx
@@ -1,0 +1,35 @@
+"use client"
+import { useScrollSectionGroup } from "@/components/ui/scroll-section-group/use-scroll-section-group";
+import { useSearchParams } from "next/navigation";
+import React, { useEffect } from "react";
+import { redesignedArbovirusPageSections, sectionUrlParameterToSectionId } from "./sections";
+
+export default function RedesignedArbovirusPage() {
+  const searchParams = useSearchParams();
+
+  const {renderScrollSectionGroup, moveScrollSectionGroupToSection} = useScrollSectionGroup({
+    scrollSectionGroupProps: {
+      sections: redesignedArbovirusPageSections,
+      className:"overflow-y-scroll col-span-10 h-full row-span-2 snap-y scroll-smooth",
+      scrollThrottleThresholdMilliseconds: 700
+    }
+  })
+
+  useEffect(() => {
+    const sectionUrlParameter = searchParams.get('section');
+    const sectionId = sectionUrlParameterToSectionId({urlParam: sectionUrlParameter})
+
+    if(sectionId) {
+      moveScrollSectionGroupToSection({ sectionId });
+    }
+  }, [searchParams])
+  
+  return (
+    <div className="col-span-12 row-span-2 grid gap-0 grid-cols-12 grid-rows-2 grid-flow-col w-screen overflow-hidden border-box -my-4 -ml-4">
+      <div className="col-span-2 h-full row-span-2">
+        <div className="w-full h-full bg-orange-500"> placeholder for filters </div>
+      </div>
+      {renderScrollSectionGroup()}
+    </div>
+  )
+}
--- a/app/pathogen/arbovirus/wip/page.tsx
+++ b/app/pathogen/arbovirus/wip/page.tsx
@@ -10,7 +10,7 @@ export default function RedesignedArbovirusPage() {
   const {renderScrollSectionGroup, moveScrollSectionGroupToSection} = useScrollSectionGroup({
     scrollSectionGroupProps: {
       sections: redesignedArbovirusPageSections,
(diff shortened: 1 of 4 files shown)


app/pathogen/arbovirus/wip/page.tsx
```diff
@@ -0,0 +1,39 @@
+"use client"
```
SeanKennyNF, line 1: The path I'm using is `http://localhost:3000/pathogen/arbovirus/wip` because this is a work in progress. I'd really love to move this to `http://localhost:3000/pathogen/arbovirus/dashboard` when we're ready to enable the feature flag but that route is obviously taken.

Anyways, if anyone has any route ideas you can just drop them here. I'm not overly concerned with how we'll do the whole route thing, we can move these files on a moments notice and nobody in production will see them right now :)

app/pathogen/arbovirus/wip/page.tsx
```diff
@@ -0,0 +1,39 @@
...
+    const sectionId = sectionUrlParameterToSectionId({urlParam: sectionUrlParameter})
+
+    if(sectionId) {
+      moveScrollSectionGroupToSection({ sectionId });
+    }
+  }, [searchParams])
+
+  if(process.env.NEXT_PUBLIC_WEBSITE_REDESIGN_ENABLED !== 'true') {
+    return notFound();
+  }
```
SeanKennyNF, line 29: The `NEXT_PUBLIC_...` in front of the variable name is non-negotiable apparently if we want to use it in client side files (which kind of makes sense). This feature flag is just so that people in production don't see my colourful page.

@HimanshuRanka if you could please set this environment variable to `true` in the preview environment but to anything that isn't `true` in the production environment just so that people can see my page on their vercel previews I would really appreciate it :heart:

app/pathogen/arbovirus/wip/page.tsx
```diff
@@ -0,0 +1,39 @@
...
+  
+  return (
+    <div className="col-span-12 row-span-2 grid gap-0 grid-cols-12 grid-rows-2 grid-flow-col w-screen overflow-hidden border-box -my-4 -ml-4">
```
SeanKennyNF, line 32: Silly negative margin thing to fight the layout. Yeah I know it's silly, it's the simplest solution to the padding on the layout please bear with me lol

app/pathogen/arbovirus/wip/sections.tsx
```diff
@@ -0,0 +1,73 @@
...
+  MAP = 'MAP',
+  TABLE = 'TABLE',
+  VISUALIZATIONS = 'VISUALIZATIONS'
+}
+
+enum RedesignedArbovirusPageSectionUrlParam {
+  map = 'map',
+  table = 'table',
+  visualizations = 'visualizations'
+}
```
SeanKennyNF, line 13: Same logic as the separation for the visualizations, we should never delete url params so that we won't break links that are out in the wild.

app/pathogen/arbovirus/wip/sections.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export const redesignedArbovirusPageSections = [{
+  id: RedesignedArbovirusPageSectionId.MAP,
+  url: RedesignedArbovirusPageSectionUrlParam.map,
+  renderScrollSectionContent: (input: { ref: RefObject<HTMLElement>, key: string}) => (
+    <section
+      key={input.key}
+      ref={input.ref}
+      className="w-full h-[95%] bg-blue-500 snap-start snap-always scroll-smooth"
```
SeanKennyNF, line 25: What I love so much about the `h-[95%]` is that if it's not the last section, the next section peeks up below the page a little bit to let you know that there's another section and you can scroll down. I think this will look really nice hopefully once the sections aren't just coloured rectangles lol

Peek 2024-01-21 16-23

app/pathogen/arbovirus/wip/sections.tsx
```diff
@@ -0,0 +1,70 @@
...
+      placeholder for visualizations
+    </section>
+  )
+}]
+
+const sectionUrlParamToSectionIdMap = {
+  [RedesignedArbovirusPageSectionUrlParam.map]: RedesignedArbovirusPageSectionId.MAP,
+  [RedesignedArbovirusPageSectionUrlParam.table]: RedesignedArbovirusPageSectionId.TABLE,
+  [RedesignedArbovirusPageSectionUrlParam.visualizations]: RedesignedArbovirusPageSectionId.VISUALIZATIONS,
+}
```
SeanKennyNF, line 58: Could be not 1-1 in the future if we change the names of the sections or want to change the url params.

components/ui/scroll-section-group/scroll-section-group.tsx
```diff
@@ -0,0 +1,83 @@
...
+  
+  const changeViewedSectionBasedOnScroll = (currentIndex: number, scrollDirection: ScrollDirection) => {
+    const nextIndex = scrollDirection === ScrollDirection.UP ? currentIndex - 1 : currentIndex + 1;
+
+    if(nextIndex < 0 || nextIndex >= props.sections.length) {
+      return;
+    }
+
+    props.setCurrentIndex(nextIndex);
+  }
```
SeanKennyNF, line 39: Yo Sean why are we writing our own JS for scroll snapping? Isn't there a CSS solution?

Yeah there is. Unfortunately,
- From my testing the CSS solution sucks if you have large scroll snap panels and it was only working like half the time anyways. Apparently it has issues on certain browsers as well so :sparkles: _no thanks_ :sparkles:. I kept the classes like `snap-always` and `smooth-scroll` on the elements since my JS scroll can still technically break if you go crazy with the scrollbar so the CSS is more of a fallback here.
- I don't think there's a way to get both the CSS snapping and my custom route based snapping (see later comment) to play nice with each other since one would be in CSS and the other would 100% have to be in JS
- I tried some of the JS libraries for scroll snapping and I didn't like any of them so this is all homebrewed and wayyyyy better in my opinion.

I'd just try it out for yourself. Honestly, every nice looking scroll panel website I saw was either not using the CSS scroll snap at all and using pure JS or is doing what I'm doing where they are using a combination of both.


Two really nice ones I'm in love with :heart:
https://www.tesla.com/ <- CSS but clearly also using JS on top of that because I know the CSS stuff doesn't work like that
https://whereispoland.com/ <- Pure JS if you check out the DOM as you scroll

Peek 2024-01-21 16-23

app/pathogen/arbovirus/wip/page.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { redesignedArbovirusPageSections, sectionUrlParameterToSectionId } from "./sections";
+
+export default function RedesignedArbovirusPage() {
+  const searchParams = useSearchParams();
+
+  const { renderScrollSectionGroup, moveScrollSectionGroupToSection } = useScrollSectionGroup({
+    scrollSectionGroupProps: {
+      sections: redesignedArbovirusPageSections,
+      className:"col-span-10 row-span-2",
+      scrollThrottleThresholdMilliseconds: 700
```
SeanKennyNF, line 14: Magic number but basically got this experimentally. If it's too low, you scroll multiple panels at once. If it's too high, you're allowed to scroll freely for a bit after changing panels which I tried to minimize to the best of my ability

400 - way too low
1000 - way too high
700 - works okay

Peek 2024-01-21 16-23

components/ui/scroll-section-group/use-scroll-section-group.tsx
```diff
@@ -0,0 +1,66 @@
...
+      hasEnoughTimePassedSinceLastScrollEventActioned={hasEnoughTimePassedSinceLastScrollEventActioned}
+      setLastScrollEventActionedUnixEpochTimestampMilliseconds={setLastScrollEventActionedUnixEpochTimestampMilliseconds}
+      scrollThrottleThresholdMilliseconds={input.scrollSectionGroupProps.scrollThrottleThresholdMilliseconds}
+      sections={input.scrollSectionGroupProps.sections.map((section, index) => ({...section, ref:sectionRefs.current[index]}))}
+      currentIndex={currentIndex}
+      setCurrentIndex={setCurrentIndex}
+    />
+  }, [input.scrollSectionGroupProps, currentIndex, setCurrentIndex]);
+
+  const moveScrollSectionGroupToSection = useCallback(({ sectionId }: MoveScrollSectionGroupToSectionInput<TSectionId>) => {
```
SeanKennyNF, line 51: Basically what I wanted here was the ability to route someone to `http://localhost:3000/pathogen/arbovirus/wip?section=map` or `http://localhost:3000/pathogen/arbovirus/wip?section=table` or `http://localhost:3000/pathogen/arbovirus/wip?section=visualizations` and have those routes immediately scroll you to the right section. If you look at the designs, there are a whole bunch of links to specific sections so I'm not making this up lol 

**I'm not scrolling at all in the gif below, the route parameter is taking me to the right spot on it's own**

Peek 2024-01-21 16-28

components/ui/scroll-section-group/scroll-section-group.tsx
SeanKennyNF: Oh yeah I also want to do some conditionals based on which page you're currently scrolled to which is a massive pain in the rear if I did this with CSS

app/pathogen/arbovirus/wip/page.tsx
SeanKennyNF: Actually I turn this to 1000 later because it was driving me nuts when it would scroll multiple panels at once.

EliasPinno (approved):

PR #155, opened 2024-02-03
Changed the graph shown on the dashboard to the "Estimate count by WHO region and arbovirus" graph.

### LINKED ISSUE #145: Change the graph displayed next to the dashboard to the "Estimate Count by WHO Region and Arbovirus" graph
From Harriet [here](https://3.basecamp.com/5061858/buckets/30589330/messages/6958342546#__recording_6962615947),

> Graph positions:
> Based on Heidelberg feedback, let's reposition the graphs as follows
> 1. Estimate count by WHO region and arbovirus (this should also be the one featured on the dashboard page for now)
> 2. Median seroprevalence of arboviruses by WHO Region
> 3. Change in median seroprevalence over time
> 4. Estimate count by arbovirus & antibody type
> 5. Cumulative estimate count over time by sample frame
> 6. Median seroprevalence by WHO region and age group
> **Remove Cumulative estimate count over time by arbovirus and Top ten countries reporting estimates by arbovirus

Let's change the graph displayed next to the dashboard to the "Estimate Count by WHO Region and Arbovirus" graph. One potential consideration would be that if you switch them as-is, the graph will look really streteched so maybe we move the legend under the graph for this particular view? Maybe we just make the graph a bit shorter and not bother with something complicated because the redesign will do away with any kind of complication here.

SeanKennyNF:
Resolves #145.
Had to do some extra stuff to move the legend to the bottom because it looked very strange with the legend to the right. The graph still looks very tall but oh well

Screenshot from 2024-02-03 11-48-28
Screenshot from 2024-02-03 11-48-43

### DIFF
--- a/app/pathogen/arbovirus/dashboard/page.tsx
+++ b/app/pathogen/arbovirus/dashboard/page.tsx
@@ -4,7 +4,7 @@ import React, { useCallback, useMemo } from "react";
 import { CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 
 import { ArbovirusMap } from "@/app/pathogen/arbovirus/dashboard/(map)/ArbovirusMap";
-import { MedianSeroPrevByWHOregion } from "../analyze/recharts";
+import { LegendConfiguration, MedianSeroPrevByWHOregion, WHORegionAndArbovirusBar } from "../analyze/recharts";
 import { FilterableField, Filters } from "./filters";
 import { CardConfiguration, CardStyle, CardType, getConfigurationForCard } from "@/components/ui/card-collection/card-collection-types";
 import { CardCollection } from "@/components/ui/card-collection/card-collection";
@@ -35,7 +35,7 @@ export default function ArbovirusDashboard() {
             <ZoomIn />
           </button>
         </div>
-        {visualizationInformation.renderVisualization()}
+        <WHORegionAndArbovirusBar legendConfiguration={LegendConfiguration.BOTTOM_ALIGNED} />
       </CardContent>
     );
   }, [router])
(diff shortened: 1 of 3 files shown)


app/pathogen/arbovirus/dashboard/page.tsx
```diff
@@ -18,22 +18,27 @@ export default function ArbovirusDashboard() {
...
+            {visualizationInformation.displayName}
+          </h3>
+          <button 
+            onClick={() => router.push(`visualizations?visualization=${getVisualizationInformationFromVisualizationId({visualizationId: VisualizationId.MEDIAN_SEROPREVALENCE_BY_WHO_REGION}).urlParameter}&referrerRoute=/pathogen/arbovirus/dashboard`)}
+            aria-label="See visualization in fullscreen"
+          >
+            <ZoomIn />
+          </button>
+        </div>
+        <WHORegionAndArbovirusBar legendConfiguration={LegendConfiguration.BOTTOM_ALIGNED} />
```
SeanKennyNF, line 38: The reason why I can't use `renderVisualizations` is because I don't want to pass in `legendConfiguration` for every single visualization since it won't have any effect on some visualizations. We're tearing down the dashboard tab soon so w/e. Just seems like a lot of work for something we'll throw away in a month so yeah lol

app/pathogen/arbovirus/visualizations/visualizations.tsx
```diff
@@ -97,7 +97,7 @@ const allVisualizationInformation: Record<VisualizationId, VisualizationInformat
       VisualizationUrlParameter["estimate-count-by-who-region-and-arbovirus"],
     displayName:
       VisualizationDisplayName["Estimate count by WHO region and arbovirus"],
-    renderVisualization: WHORegionAndArbovirusBar,
+    renderVisualization: () => WHORegionAndArbovirusBar({ legendConfiguration: LegendConfiguration.RIGHT_ALIGNED }),
```
SeanKennyNF, line 100: Yeah silly magic but we'll throw this away soon.

PR #157, opened 2024-02-04
Added country pop-up modal

### LINKED ISSUE #108: Add country information popup when a country is clicked on.
> Show Number of studies and then seroprev per country (add hover over country metadata like in serotracker.com)

Below is a screenshot of what a country information modal looks like on https://serotracker.com. I'm thinking we'd have to throw the popup in `PathogenCountryHighlightLayer` honestly, that's where the country vectors live and clicking on a country would click on that layer so I think this popup has to live in that layer also. I'm thinking we could fetch the number of estimates for each pathogen and the median seroprevalence of each pathogen per country. Let's maybe start with that and then if we want more information let's hit that as part of a separate ticket.

image

SeanKennyNF:
Resolves #108. See tickie for more details.

Peek 2024-02-03 22-44

image

### DIFF
--- a/app/pathogen/arbovirus/analyze/recharts.tsx
+++ b/app/pathogen/arbovirus/analyze/recharts.tsx
@@ -25,6 +25,15 @@ import clsx from "clsx";
 
 //Study count by pathogen and antibody type
 
+export enum ShortformArbovirus {
+  DENV = "DENV",
+  ZIKV = "ZIKV",
+  CHIKV = "CHIKV",
+  YF = "YF",
+  WNV = "WNV",
+  MAYV = "MAYV",
+}
+
 export type arbovirusesSF = "DENV" | "ZIKV" | "CHIKV" | "YF" | "WNV" | "MAYV";
 type arboviruses =
   | "Dengue"
--- a/app/pathogen/arbovirus/dashboard/(map)/ArbovirusMap.tsx
+++ b/app/pathogen/arbovirus/dashboard/(map)/ArbovirusMap.tsx
@@ -16,6 +16,7 @@ import { PathogenMap } from "@/components/ui/pathogen-map/pathogen-map";
 import { MapArbovirusFilter } from "./MapArbovirusFilter";
 import { MapExpandPlotsPrompt } from "./MapExpandPlotsPrompt";
 import { MapArbovirusStudySubmissionPrompt } from "./MapArbovirusStudySubmissionPrompt";
+import { ArboCountryPopupContent } from "../ArboCountryPopUpContent";
 
 export const pathogenColorsTailwind: { [key: string]: string } = {
   ZIKV: "border-zikv data-[state=checked]:bg-zikv",
@@ -97,9 +98,13 @@ export function ArbovirusMap(props: ArbovirusMapProps) {
               },
             },
           ]}
-          generatePopupContent={(record) => (
-            <ArboStudyPopupContent record={record} />
-          )}
+          generatePopupContent={(input) => {
+            if(input.layerId === 'country-highlight-layer') {
+              return <ArboCountryPopupContent record={input.data} />
+            }
+          
+            return <ArboStudyPopupContent record={input.data} />
+          }}
         />
       </div>
       <MapArbovirusStudySubmissionPrompt 
--- a/app/pathogen/arbovirus/dashboard/ArboCountryPopUpContent.tsx
+++ b/app/pathogen/arbovirus/dashboard/ArboCountryPopUpContent.tsx
@@ -1,0 +1,8 @@
+
+interface ArboCountryPopupContentProps {
+  record: any;
+}
+
+export function ArboCountryPopupContent({ record }: ArboCountryPopupContentProps): React.ReactNode {
+  return <div> {record.countryName} </div>
+}
--- a/app/pathogen/arbovirus/dashboard/ArboCountryPopUpContent.tsx
+++ b/app/pathogen/arbovirus/dashboard/ArboCountryPopUpContent.tsx
@@ -1,8 +1,49 @@
+import { typedGroupBy } from "@/lib/utils"
+import { ShortformArbovirus, convertArboSFtoArbo } from "../analyze/recharts"
+import { PopUpContentRow } from "./ArboStudyPopupContent";
 
 interface ArboCountryPopupContentProps {
-  record: any;
+  record: {
+    id: string,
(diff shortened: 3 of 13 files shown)


app/pathogen/arbovirus/analyze/recharts.tsx
```diff
@@ -25,6 +25,15 @@ import clsx from "clsx";
 
 //Study count by pathogen and antibody type
 
+export enum ShortformArbovirus {
```
SeanKennyNF, line 28: I'd like to move us towards an enum. The big advantage is that I can `Object.values` it to get every enum value which I make use of in this PR. The codegen library I want to use here also uses enums instead of `'X' | 'Y' | 'Z'` so don't worry about that :)

app/pathogen/arbovirus/dashboard/(map)/ArbovirusMap.tsx
```diff
@@ -97,9 +98,13 @@ export function ArbovirusMap(props: ArbovirusMapProps) {
               },
             },
           ]}
-          generatePopupContent={(record) => (
-            <ArboStudyPopupContent record={record} />
-          )}
+          generatePopupContent={(input) => {
+            if(input.layerId === 'country-highlight-layer') {
```
SeanKennyNF, line 102: Hardcoded string kind of silly but will have to do for now I suppose. Might take a second look at maybe something we can do later when my TODO list is a bit thinner.

app/pathogen/arbovirus/dashboard/ArboCountryPopUpContent.tsx
```diff
@@ -0,0 +1,49 @@
...
+    longitude: string,
+    dataPoints: { pathogen: ShortformArbovirus }[],
+  }
+}
+
+export function ArboCountryPopupContent({ record }: ArboCountryPopupContentProps): React.ReactNode {
+  const dataPointsGroupedByArbovirus = typedGroupBy(record.dataPoints, (dataPoint) => dataPoint.pathogen);
+
+  return (
+    <div className="w-[460px] bg-white pt-2">
```
SeanKennyNF, line 20: There's a bit of copy-pastel from `ArboStudyPopUpContent` like the maximum width, the paddings, stuff like that. Probably worth consolidating some of that stuff into components, just not right now.

app/pathogen/sarscov2/dashboard/(map)/MapAndFilters.tsx
```diff
@@ -95,9 +95,13 @@ export default function MapAndFilters() {
                   },
                 },
               ]}
-              generatePopupContent={(record) => (
-                <SarsCov2StudyPopupContent record={record} />
-              )}
+              generatePopupContent={(input) => {
+                if(input.layerId === 'country-highlight-layer') {
+                  throw new Error("The SARS CoV2 Country pop-up is unimplemented");
+                }
```
SeanKennyNF, line 101: Attaining maximum laziness level lmao

In all seriousness, we'll come back for this when we move the CoV2 stuff over hahaha
It's feature flagged off so we're 100% cool. I'm just doing this because the types are `any` and it isn't clear you have to do this if I don't explicitly write it like this.

lib/country-iso-3166-1-alpha-3-codes.ts
```diff
@@ -193,3 +195,7 @@ export const countryNameToIso31661Alpha3CodeMap: Record<string, string> = {
   "St Lucia": "LCA",
   "Saint Lucia": "LCA",
 };
+
+export const iso31661Alpha3CodeToCountryNameMap: Record<string, string> = typedObjectFromEntries(
+  typedObjectEntries(countryNameToIso31661Alpha3CodeMap).map(([key, value]) => [value, key])
```
SeanKennyNF, line 200: If you're curious what name it uses if there's multiple entries with the same alpha 3 country code, it uses the last one in the list. Pretty minor detail, I want to move all of this stuff to the backend anyways.

lib/bounding-boxes.ts
```diff
@@ -26,6 +26,15 @@ export const combineBoundingBoxes = (
   );
 };
 
+export const getBoundingBoxCenter = (boundingBox: BoundingBox): {latitude: number, longitude: number} => {
+  const [minLng, minLat, maxLng, maxLat] = boundingBox;
+
+  return {
+    latitude: (maxLat + minLat) / 2,
+    longitude: (maxLng + minLng) / 2
```
SeanKennyNF, line 34: On very rare occasions (Malaysia) this will be in the water but I swear it really doesn't bother me and it's really not obvious.

app/pathogen/arbovirus/analyze/recharts.tsx
HimanshuRanka, line 28: <3

app/pathogen/arbovirus/dashboard/ArboStudyPopupContent.tsx
```diff
@@ -2,18 +2,16 @@ import { TranslateDate } from "@/utils/translate-util/translate-service";
...
- * @param content: right content of study modal: the actual content for this record
- */
-function row(
+interface PopUpContentRowProps {
   title: string,
   content: JSX.Element | string | string[] | null | undefined
-) {
+}
+
+export const PopUpContentRow = (props: PopUpContentRowProps): React.ReactNode => {
```
HimanshuRanka, line 10: ++++1

components/ui/pathogen-map/pathogen-map.tsx
```diff
@@ -42,9 +46,28 @@ export function PathogenMap<
   generatePopupContent,
   layers,
 }: PathogenMapProps<TPathogenDataPointProperties>) {
-  const [popUpInfo, setPopUpInfo] = useState<
+  const [popUpInfo, _setPopUpInfo] = useState<
```
HimanshuRanka, line 49: why the underscore?

HimanshuRanka (approved):
LGTM

components/ui/pathogen-map/pathogen-map.tsx
SeanKennyNF: Ahhhh it is denoting that it's private but I want to give it the same name. People do this sometimes to tell people that I want to call `setPopUpInfo` everywhere instead of `_setPopUpInfo`, I think I saw this on some Kotlin tutorial or something lol

Issue #132, opened 2024-01-14
Implement the new "About" page


SeanKennyNF:
This ticket is to implement the new "About" page from the Figma designs which can be found [here](https://www.figma.com/file/BKtEVRMNkZYh4EbG9T5wp5/PathoTracker?type=design&node-id=0-1&mode=design)

Couple notes:
- https://serotracker.com/en/About and https://serotracker.com/en/Data#FAQ share some components in common with the new designs and it's likely that those pages in [this repo](https://github.com/serotracker/sero-can-webapp) have some code that can be copy-pasted over with not too much trouble
- Quote from https://github.com/PathoTracker/Pathotracker/issues/21, _"The data on the about page at the moment is highly outdated and irrelevant in a lot of ways and places. This needs to be updated under guidance from Mairead, Harriet and Sabah"_

Relevant designs:
![image](https://github.com/PathoTracker/Pathotracker/assets/86806388/03347912-bea4-45f0-b3f8-4b7310728db5)
![image](https://github.com/PathoTracker/Pathotracker/assets/86806388/a4b7dcda-8386-42a8-827b-67de7527cc51)
![image](https://github.com/PathoTracker/Pathotracker/assets/86806388/72bb2973-c28e-45d4-9f3d-29b718ce7ce7)

SeanKennyNF:
This page was added, closing issue.

PR #172, opened 2024-02-11
Added navigation and a sidebar for the new about page layout

SeanKennyNF:
One of the steps in resolving #132 

This just adds a skeleton page which I'm going to add content to.

Peek 2024-02-10 20-39


app/about/about-page-context.tsx
```diff
@@ -0,0 +1,61 @@
...
+export const AboutPageProvider = ({
+  children,
+}: {
+  children: React.ReactNode;
+}): React.ReactNode => {
+  const queryClient = new QueryClient({
+    defaultOptions: {
+      queries: {
+        staleTime: 86400000, // One day
+        cacheTime: 86400000, // One day
```
SeanKennyNF, line 49: I've got big plans for refreshing the data for this page daily but this won't make sense until that's done hahaha

app/about/data-extraction/page.tsx
```diff
@@ -0,0 +1,16 @@
+"use client";
```
SeanKennyNF, line 1: These are intentionally empty for now, this is all behind a feature flag so I'm not making us look bad hahaha

app/about/page.tsx
SeanKennyNF, line 1: This page was inaccessible unless you put had the url and was super out of date anyways. I didn't really see a point to keeping it around.

hooks/useGroupedTeamMemberData.tsx
```diff
@@ -0,0 +1,28 @@
+import { useQuery } from "@tanstack/react-query";
+import { gql } from "@apollo/client";
+import { request } from 'graphql-request';
+
+export const groupedTeamMembersQuery = gql`
+  query groupedTeamMembersQuery {
```
SeanKennyNF, line 6: I'm going to use this to populate the "The Team" page. I'll add some types then.

app/about/about-page-context.tsx
```diff
@@ -0,0 +1,61 @@
+"use client";
```
HimanshuRanka, line 1: Hmmm so my thoughts on this are that it may be overkill. The about page in general is mainly informational and I am not sure we need to have a context for it (I assume this is to save their last visted/currentPage?) - I see the value in that conceptually but I wonder if there is even a use case for it.

If I am missing something else here, let me know.

app/about/about-page-context.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export const AboutPageProvider = ({
+  children,
+}: {
+  children: React.ReactNode;
+}): React.ReactNode => {
+  const queryClient = new QueryClient({
+    defaultOptions: {
+      queries: {
+        staleTime: 86400000, // One day
```
HimanshuRanka, line 48: I see the value in this but our stale time could be much longer also/could be infinte (the cache time and stale time only matter for current session afaik - might be worng but this is my understanding so please correct me if not) Our team data will not change often enough to use caching at all. infact the page will not be accessed often enough either so we could just call it everytime (maybe to prevent spam we can add stale time) anyway all this is just discussion and I feel like I have come to a point where a 1 day long stale time and cache time is good for preventing someone from spamming our servers accidentally even (maybe overkill too ahahaha)

app/about/about-page-context.tsx
HimanshuRanka: If I have missed something above please let me know!

app/about/data-extraction/page.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { useContext, useEffect } from "react";
+import {
+  AboutPageSidebarOption,
+  aboutPageSidebarContext,
+} from "../about-page-context";
+
+export default function DataExtractionPage() {
+  const { setCurrentSidebarOption } = useContext(aboutPageSidebarContext);
+
+  useEffect(() => {
```
HimanshuRanka, line 11: For this and all the following pages, why do we have a useEffect? what is this a side effect of?

IF this is to set the context when a user navigates to this page, then IMO that should happen when the button is clicked not when this page loads.

HimanshuRanka (changes requested):
Mostly looks good to me! I think outside of the context and related effects it all makes sense to me 😄

SeanKennyNF (commented):
@HimanshuRanka thanks for making me take a second look, I have no idea what I was thinking with some of this lol. I've made a revised version for you to take a peek at your next convenience :)

- No more `useEffect`s, I don't know why I thought that was the right way to get what I wanted lol
- No query caching fanciness. I agree that it's putting the cart before the horse. Let's just make the query every time the page is loaded and once the page is functional we can look at whether we want infinite cache time or not.
- No more context, that tied into the `useEffect`s and again no idea what I was thinking lol

app/about/about-page-context.tsx
```diff
@@ -0,0 +1,23 @@
...
+    defaultOptions: {
+      queries: {
+        staleTime: Infinity,
+        cacheTime: Infinity
+      },
+    },
+  });
+  
+  return (
+    <QueryClientProvider client={queryClient}>
```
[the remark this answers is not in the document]
SeanKennyNF: Yes we do :'(

The other query client provider lives at the base of the layout in the `arbovirus` route so if this were `arbovirus/about/the-team` I'd be set. I think there's potentially value in making some kind of root layout that has this provider, maybe a tiny tech debt ticket?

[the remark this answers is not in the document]
SeanKennyNF: I'm going to go ahead and merge this but if you're on board for trying one layout with providers for every page, let's make a tiny ticket for that :)

PR #175, opened 2024-02-11
Added a basic error and not found page which is routed to correctly.

EliasPinno:


app/error.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { useEffect } from 'react'
+ 
+export default function Error({
+  error,
+  reset,
+}: {
+  error: Error & { digest?: string }
+  reset: () => void
+}) {
+  useEffect(() => {
```
HimanshuRanka, line 13: Unsure if we need this useEffect here (irrespective of vercel logging the error) or not --> u could just console.log(error) outside the effect and it would still print everytime the error changes as it would cause this component to rerender since error comes in as a prop.

HimanshuRanka: Stupid rule i use that is super handy --> If you are using a useEffect, don't.

There will come a point where you will need it (which is why it exists) but 99% of the time there is a better way to do it. Reading reacts own documentation on useEffect gave me a really goo dunderstanding on this.

app/error.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <div className='flex justify-center items-center h-screen'>
+        <div className='prose pb-2'>
+            <div className='flex justify-center'>
+                <h1 className="mb-0">Oops!</h1>
+            </div>
+            <p>Sorry: The website encountered an error trying that operation! It returned this message:</p>
+            <div style={{ color: '#ff0000', backgroundColor: '#f5f5f5' }} className='rounded'>{error.message}</div>
+            <p>If you want that operation again, click below!</p>
+            <div className='w-full flex justify-center'>
+                <Button className="" onClick={() => reset()}>
```
HimanshuRanka, line 27: Can we add the tailwind class bg-background to this button --> will make the button the same color as the header (background is a cutom variable we have set in the config)

also in another siutation if we a prop is empty just remove it (classname is almost never required and if it is that should indicate an issue with the component structure)

app/not-found.tsx
```diff
@@ -0,0 +1,25 @@
...
+    const currentPage = usePathname();
+    return (
+    <div className='flex justify-center items-center h-screen'>
+        <div className='prose pb-2'>
+            <div className='flex justify-center'>
+                <h1 className="mb-0">404: Page Not Found</h1>
+            </div>
+            <p>Sorry, but the route <span className='p-1' style={{ color: '#ff0000', backgroundColor: '#f5f5f5' }}>{currentPage}</span> could not be found. Click the button below to navigate back home.</p>
+            <div className='w-full flex justify-center'>
+                <Button className="" onClick={() => {router.push("/")}}>
```
HimanshuRanka, line 16: Same as above

app/not-found.tsx
```diff
@@ -0,0 +1,25 @@
...
+                    Back to home
```
HimanshuRanka, line 17: Also instead of back to home can navigate back to previous page

HimanshuRanka (changes requested):

app/not-found.tsx
EliasPinno: Done.

app/error.tsx
EliasPinno: Done.

app/not-found.tsx
EliasPinno: Done.

HimanshuRanka (dismissed):

SeanKennyNF (dismissed):
:fire:

Reminder to remove `test_error/page.tsx` before merging

SeanKennyNF (approved):

PR #177, opened 2024-02-11
Created the "The Team" Page

SeanKennyNF:
Created the "The Team" page.

Peek 2024-02-11 13-22


hooks/useGroupedTeamMemberData.tsx
```diff
@@ -20,8 +20,24 @@ export const groupedTeamMembersQuery = gql`
   }
 `
 
+interface GroupedTeamMembersQuery {
```
HimanshuRanka, line 23: ❓ So did this come from gql/mongo and can it be dynamic, or did we copy it (still kinda confused on how the mongo gql typescript typing integration works)

app/about/the-team/page.tsx
```diff
@@ -4,13 +4,88 @@ import {
   AboutPageSidebarOption,
```
HimanshuRanka, line 4: Sorry if this is a premature review and you are already working on all of this ahahahahah but I just wanted to point that out really quick. 

Right off the bar I think there are a lot of stylistic improvements that need to be made here. 
- The font size is off, I think it is too large throughout
- We want to add padding to the whole page (this might be a layout thing) and increase the white space around the content of the pages and the navigational components (sidebar, header, footer) : There is a very specific directive on the width of a page that should have text content. If you look up UI rules for blogs etc it will tell u what percentage of the view width should have content. That is something we want to take into consideration. Sometimes the suggested percentage is too low on case by case basis (in our case it would be since we already have text only in a smaller portion of the view width) but that is something that we can settle on using trial and error and feel of the page (does it seem too overwhelming, is it too large)
- We want to make all the logos the actual logos rather than logo based art (blue bird for twitter - now x but i like the bird, the blue box with white text for linked in and black solid filled in envelop for email) they can be a lot smaller and maybe the area around them goes gray when u hover over them. The logos can also be smaller
- the space between each section (team leads, dev team) etc. should be larger than the space between header and content (team leads/ mairead, harriet, nik)

I would say in general the current style of the old about page is really good and we want to replicate that pretty much to the T. So use that as reference.

app/about/the-team/page.tsx
```diff
@@ -4,13 +4,88 @@ import {
...
+const TeamMemberInfoCard = (props: TeamMemberInfoCardProps) => {
+  return (
+    <div> 
+      <b> {`${props.teamMember.firstName} ${props.teamMember.lastName}`} </b>
+      {props.teamMember.affiliations.map((affiliation) => 
+        <p key={`${props.teamMember.firstName}-${props.teamMember.lastName}-${affiliation.label}`}> {affiliation.label} </p>
+      )}
+      <div className="flex flex-row mt-2">
+        {props.teamMember.linkedinUrl ? (
+        <a className="pr-2 text-link" href={props.teamMember.linkedinUrl} target="_blank" rel="noopener noreferrer">
```
HimanshuRanka, line 39: using a vs using next/Link - do we know which is better and what to use when?

HimanshuRanka: Is probably also the thing making the logos blue.

app/about/the-team/page.tsx
SeanKennyNF: Not at all man, I appreciate the early look :)

> The font size is off, I think it is too large throughout

Yes you're right it's slightly larger. The "Our Team" header is font-size 36px which is the closest I can get to the old website's 34.8px (30px too small, 48px to large). The " Principal Investigator", "Evidence Synthesis and Data Curation Team", etc. have a font size of 30px which is the closest I can get to the old website's 29px. Would you prefer this? It downsizes the fonts by one tailwind size everywhere but now it's a fair amount smaller than the old site's.

image

> We want to add padding to the whole page (this might be a layout thing) and increase the white space around the content of the pages and the navigational components (sidebar, header, footer) : There is a very specific directive on the width of a page that should have text content. If you look up UI rules for blogs etc it will tell u what percentage of the view width should have content. That is something we want to take into consideration. Sometimes the suggested percentage is too low on case by case basis (in our case it would be since we already have text only in a smaller portion of the view width) but that is something that we can settle on using trial and error and feel of the page (does it seem too overwhelming, is it too large)

Would it be fair if we shipped all of this and I opened a separate PR with this? I would definitely need to do some reading and don't want to block on this. You're right that this would be a layout thing I feel.

> We want to make all the logos the actual logos rather than logo based art (blue bird for twitter - now x but i like the bird, the blue box with white text for linked in and black solid filled in envelop for email) they can be a lot smaller and maybe the area around them goes gray when u hover over them. The logos can also be smaller

Yeah for sure man, I changed some stuff here so you might like the newer version! Let me know if you have any feedback though! I did use fontawesome because lucide doesn't have brand logos besides the ones you've seen. I could also just download the svgs if we don't want to use the library. Your call :)

> the space between each section (team leads, dev team) etc. should be larger than the space between header and content (team leads/ mairead, harriet, nik)

I just shipped a change to make it more like the dimensions from the original site so please let me know what you think :)

hooks/useGroupedTeamMemberData.tsx
SeanKennyNF: I wish it came from gql :sob: 
The end plan is to have these autogenerated but I generated this by hand from the gql schema.

HimanshuRanka: LOL and what do we as devs need to do to generate stuff from gql?

HimanshuRanka:
Overall think the changes made a huge difference but there are a bunch of layout concerns we will need to address soon! and yes I am ok with opening a seperate PR for that (and ticket ofc)

hooks/useGroupedTeamMemberData.tsx
SeanKennyNF: Ahhhhh that one's with @EliasPinno right now.

This is a tutorial for the most basic graphql front end typegen setup [link](https://the-guild.dev/graphql/codegen/docs/getting-started). Just the front end part, you can disregard the backend part because that's already set up. Basically you the library downloads the graphql schema and then it generates types for all your different queries.

Not a big deal but once that's done I can swap out this type for the autogenerated one. I'm 100% sure the autogenerated type will be basically the exact same as this one hahaha.

app/about/the-team/page.tsx
```diff
@@ -1,7 +1,86 @@
 "use client";
+import { useGroupedTeamMemberData } from "@/hooks/useGroupedTeamMemberData";
```
SeanKennyNF, line 2: > Overall think the changes made a huge difference but there are a bunch of layout concerns we will need to address soon! and yes I am ok with opening a seperate PR for that (and ticket ofc)

Yeah that sounds good to me man, I think some kind of "design review" ticket would be good. We could get some feedback in basecamp on some of the copy on the different pages as well :)

HimanshuRanka (approved):

PR #180, opened 2024-02-12
Created the "FAQ" page

SeanKennyNF:
Created the "FAQ" page

Peek 2024-02-11 17-21

### DIFF
--- a/app/about/faq/page.tsx
+++ b/app/about/faq/page.tsx
@@ -4,6 +4,8 @@ import {
   AboutPageSidebarOption,
   aboutPageSidebarContext,
 } from "../about-page-context";
+import { Accordion } from "@/components/ui/accordion/accordion";
+import { FAQPageOptionId, faqPageText } from "./text";
 
 export default function FAQPage() {
   const { setCurrentSidebarOption } = useContext(aboutPageSidebarContext);
@@ -12,5 +14,19 @@ export default function FAQPage() {
     setCurrentSidebarOption(AboutPageSidebarOption.FAQ);
   }, [setCurrentSidebarOption]);
 
-  return <p> Faq page </p>;
+  const options = Object.values(FAQPageOptionId).map((optionId) => ({
+    id: optionId,
+    label: faqPageText[optionId].label,
+    content: faqPageText[optionId].content
+  }));
+
+  return (
+    <div className="mx-4">
+      <h1 className="text-3xl mt-5 mb-3.5"> Frequently Asked Questions </h1>
+      <Accordion
+        defaultOption={undefined}
+        options={options}
+      />
+    </div>
+  );
 }
--- a/app/about/faq/text.tsx
+++ b/app/about/faq/text.tsx
@@ -1,5 +1,4 @@
 import Link from "next/link";
-import { NextRouter } from "next/router";
 
 export enum FAQPageOptionId {
   WHERE_DOES_SEROTRACKERS_DATA_COME_FROM = 'WHERE_DOES_SEROTRACKERS_DATA_COME_FROM',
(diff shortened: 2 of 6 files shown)


app/about/faq/text.tsx
```diff
@@ -0,0 +1,37 @@
+import Link from "next/link";
```
SeanKennyNF, line 1: I don't know what about this file makes the Github coloring lose it's mind but sorry about that. All of this copy is pulled from the serotracker.com FAQ section but I removed anything that was super CoV2 related or references documents which made sense for CoV2 and didn't have a arbovirus equivalent and wouldn't make sense since we don't have CoV2 yet and added some arbovirus related language.

I'm not super concerned about the copy here, I think it's reasonable enough for now. This is behind a feature flag, we can massage the copy a bit later.

components/ui/accordion/accordion-styles.css
SeanKennyNF, line 1: I know we don't like doing this but I don't know how to do this in tailwind.

HimanshuRanka: All good, maybe we could link this weekend and try to pair program it. I might give it a shot before so I do not come in blind. lmk if yo're good with that! Also totally ok to use CSS styles cause tailwind does not offer complete css functionality in a lot of cases and this may be one of them

HimanshuRanka: Also if I am being very honest I am totally ok just using the shadcn default styling for the accordion

https://ui.shadcn.com/docs/components/accordion

components/ui/accordion/accordion.tsx
```diff
@@ -0,0 +1,40 @@
+import {
```
HimanshuRanka, line 1: Oooooo I see that u built the accordion yourself. I think going forward we should use the https://ui.shadcn.com/ library as much as possible. it gives us prebuilt components that can be fully customized but come with inbuilt and consistent style out of the box that I really like overall. It also saves us a lot of time (most of the time)

Most components in the components/ui folder are shadcn components. If we build reusable components from scratch we have been putting them in the components/custom folder. Sorry if this was not more clear ahaha

app/about/faq/text.tsx
SeanKennyNF: Yeah this is a good idea. I went back to my first PR and renamed the feature flag so there's two now :+1:

components/ui/accordion.tsx
```diff
@@ -0,0 +1,58 @@
+"use client"
```
SeanKennyNF, line 1: Autogenerated by shadcn

components/ui/accordion/accordion-styles.css
SeanKennyNF: Yeah for sure! I've switched this over to using shadcn :+1:

components/ui/accordion/accordion.tsx
SeanKennyNF: Hahahaha all good man. I've switched over to shadcn now and when I've got a spare moment I can move my custom components from ui to custom :)

HimanshuRanka: This makes me think that we should rename ui directory to something like shadcn that is more clear about what should go there cause I understand why ui and custom does not have distinct boundaries

You can also customize shadcn components (as I did in my header PR) if we need to

HimanshuRanka:
Also cant seem to link myself to the faq page on the preview, not sure why

app/about/faq/page.tsx
```diff
@@ -1,7 +1,35 @@
 "use client";
+import {
```
SeanKennyNF, line 2: > Also cant seem to link myself to the faq page on the preview, not sure why

@HimanshuRanka 

My guess would be that `NEXT_PUBLIC_ABOUT_PAGE_ENABLED` isn't set to `'true'` in the preview environment. Maybe worth taking a look?

HimanshuRanka (approved):

SeanKennyNF:
@HimanshuRanka sorry had to re-request, merge conflicts.

HimanshuRanka (approved):

Issue #158, opened 2024-02-04
Implement the redesigned header


SeanKennyNF:
This ticket is to implement the new header. Figmas are given [here](https://www.figma.com/file/BKtEVRMNkZYh4EbG9T5wp5/PathoTracker?type=design&node-id=0-1&mode=design) and some screenshots of the new header are given at the bottom of the ticket.

### A few notes

- We don't currently support any language besides English so let's either not include the language dropdown (the option labelled "EN" on the header) or let's feature flag it off.
- I feel like the "Trackers" menu in the header should expand when hovered over instead of when clicked on that's possibly just my personal opinion though
- You can link to one particular section of the new Arbotracker dashboard by using links the links `http://localhost:3000/pathogen/arbovirus/wip?section=map`, `http://localhost:3000/pathogen/arbovirus/wip?section=table`, and `http://localhost:3000/pathogen/arbovirus/wip?section=visualizations`. You'll want to use these for the "Map", "Data", and "Visualizations" sections. Basically what those links do is autoscroll to the section specified in the url parameter. SarsCoV2 tracker doesn't have a consolidated dashboard so let's just use the normal `/dashboard` and `/analyze` links for it.
- Let's hide all of the SarsCoV2 options behind the `SARS_COV_2_TRACKER_ENABLED`. You might have problems with that feature flag because it isn't prefixed with `NEXT_PUBLIC` (ie `NEXT_PUBLIC_SARS_COV_2_TRACKER_ENABLED`), that's something I've run into before so I would just change the name if you're struggling to make it work.
- Definitely hide the header behind the `NEXT_PUBLIC_WEBSITE_REDESIGN_ENABLED` feature flag. The `wip` route is also hidden behind that feature flag and we don't want to make dead links.

![image](https://github.com/PathoTracker/Pathotracker/assets/86806388/98c9234d-6548-41aa-9f19-118b3ed33e5a)

![image](https://github.com/PathoTracker/Pathotracker/assets/86806388/b971e261-a309-4d52-af09-87aaaf7314bb)

SeanKennyNF:
Closing, resolved in https://github.com/serotracker/dashboard/pull/181

Issue #133, opened 2024-01-16
Update the favicon


HimanshuRanka:
We want to replace the vercel favicon with the sero tracker logo

PR #181, opened 2024-02-18
Issue 158 redesign header

HimanshuRanka:
This PR addresses #158  and #133 

I update our header and navbar to be more in line with the design here: [figma |https://www.figma.com/file/BKtEVRMNkZYh4EbG9T5wp5/PathoTracker?type=design&node-id=310-401&mode=design&t=2fv4Vv0lqvPLoVbP-4]

The colors were inverted based on feedback given during UI review of liking how the dark color up top looked over how the white header looked. 

A lot of the changes were made on a higher level to the shadcn components themselves. This was because the likelihood that we were to use these components in another place are very low, and hence for this use case making those changes up top was great

TODO:

- [x] Make the Heading adapt to the tracker. (Sero -> SC2 -> Arbo)
- [x] Extract the styling applied to the ListItem component as well as extract the navgroup section into a custom component to standardize the styling accross the header
- [x] Add font sizing to the text
- [x] ~~Explore possiblity of having the border at the bottom move with the opening. (for this I need to gitgud at animation and I am not right now -- so will table it)~~
- [x] Update the links for each of the tabs to actually point to the right location once @SeanKennyNF merges his code


HimanshuRanka:
This is what the color changing looks like

screencast-localhost_3000-2024 02 19-14_10_47

components/ui/navigation-menu.tsx
```diff
@@ -41,7 +41,7 @@ NavigationMenuList.displayName = NavigationMenuPrimitive.List.displayName
 const NavigationMenuItem = NavigationMenuPrimitive.Item
```
HimanshuRanka, line 41: Althought this is a component from shadcn, I made a few changes here since we are 99% not going to use this comp anywhere else but the header. Even in the situation we do use it somewhere else it is likely we would want to keep these color styles and themeings.

package.json
```diff
@@ -26,7 +26,6 @@
     "@radix-ui/react-tooltip": "^1.0.7",
     "@tanstack/react-query": "^4.33.0",
     "@tanstack/react-table": "^8.9.8",
```
HimanshuRanka, line 28: This was 10$ a monthe xtra which is dumb and we do not need it as much atm.

tailwind.config.js
```diff
@@ -25,8 +25,8 @@ module.exports = {
         '16': 'repeat(16, minmax(0, 1fr))',
       },
       height: {
```
HimanshuRanka, line 27: Increased the header height to h-14 to accomodate the h1 size text. h-14 is 3.5rem and so put that here to still have fullscreen be fullscreen

SeanKennyNF (approved):
Very cool! Couple comments I think are worth addressing wrt to copy and some feature flagging issues but I do really like the new design.

components/customs/header.tsx
```diff
@@ -54,98 +55,112 @@ const arbotracker: navMenuItem[] = [
...
-                    key={page.title}
-                    title={page.title}
-                    href={page.href}
-                  >
-                    {page.description}
-                  </ListItem>
-                ))}
-              </ul>
+              <div className="p-4 flex justify-center w-full">
+                {process.env.SARS_COV_2_TRACKER_ENABLED &&
```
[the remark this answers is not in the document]
HimanshuRanka: I was wodnering whats going on
what do u mean?

components/customs/header.tsx
```diff
@@ -54,98 +55,112 @@ const arbotracker: navMenuItem[] = [
     description:
       "A collection of visualizations and tabular data tools for our collection of arbovirus data",
   },
+  {
+    title: "Visualizations",
+    href: `/pathogen/arbovirus/dashboard#${ArbovirusPageSectionId.VISUALIZATIONS}`,
+    description:
+      "A collection of visualizations and tabular data tools for our collection of arbovirus data",
```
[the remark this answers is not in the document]
HimanshuRanka: Yeah I agree, I can do that

components/customs/header.tsx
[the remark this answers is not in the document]
SeanKennyNF: Yeah fair enough I explained this bad.

NextJS stores environment variables but it isn't sure which ones are public ones that the client can use and which ones are secrets that should only be used serverside.

The way it can tell the difference is whether the environment variable starts with `NEXT_PUBLIC` or not. If it starts with `NEXT_PUBLIC`, the client is allowed to access it, if it doesn't, the client isn't. I assume that my old code was getting recognized as server side code and that's why I was getting away with `SARS_COV_2_TRACKER_ENABLED` but it looks like your code is getting rendered client side so it will always be `undefined` since it's not a public env variable.

I feel like if we change this feature flag to `NEXT_PUBLIC_SARS_COV_2_TRACKER_ENABLED` everything will start working again.

components/customs/header.tsx
```diff
@@ -1,151 +1,192 @@
...
     href: "/pathogen/sarscov2/dashboard",
     description: "A dashboard for SarsCov2 seroprevalence data",
   },
   {
-    title: "Analysis",
-    href: "/pathogen/sarscov2/analyze",
+    title: "Data",
+    href: "/pathogen/sarscov2/dashboard",
+    description:
+      "A collection of visualizations and tabular data tools for our collection of SarsCov2 data",
```
SeanKennyNF, line 36:
```suggestion
      "View or download our entire SarsCov2 dataset",
```
For consistency with the other one. The visualizations one should probably be "A collection of visualizations for our SarsCov2 dataset" as well

app/globals.css
```diff
@@ -1,92 +1,113 @@
...
+    font-size: 2rem;
+  }
+
+  h2{
+    font-size: 1.5rem;
+  }
+
+  h3 {
+    font-size: 1.25rem;
+  }
```
SeanKennyNF, line 93: I think this might have changed the font of the graphs and made the headers of the pathogens and by extension, the visualizations, noticeably larger. I preferred the old ones personally. Is there a way we can keep that font size for the graphs?

image
image

SeanKennyNF (commented):
One real quick note about what I believe to be an unintended consequence of the tailwind style changes. Also sorry for the merge conflicts again :crying_cat_face:

app/globals.css
HimanshuRanka: Oooo I did not notice that. I think we can, we would just have to specifiy the style in there if need be. let me see

app/globals.css
```diff
@@ -1,92 +1,113 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
- 
+
```
SeanKennyNF, line 4: I'm experiencing some mild visual bugs on that latest deploy.
Peek 2024-03-02 17-49

HimanshuRanka: Yeah there was a merge issue that deleted the styles, just added them back in

SeanKennyNF (commented):
Last comment I promise hahaha. A couple graphs still have `<h2>`s instead of `<p>`s so their text is appearing a bit large. I haven't converted all our visualizations to `split-bar-chart.tsx` quite yet. Could we go in an change those real quick? Looks good otherwise!

image

image

HimanshuRanka:
> Last comment I promise hahaha. A couple graphs still have `<h2>`s instead of `<p>`s so their text is appearing a bit large. I haven't converted all our visualizations to `split-bar-chart.tsx` quite yet. Could we go in an change those real quick? Looks good otherwise!
> 
> image
> 
> image

Done!

I also went ahead and made it so that the filter colors match the header color. Gonna take a look at themeing and how to do that more efficiently soon, but for now this is a good bandaid because I think the themeing will be a slightly bigger undertaking and I want to do that in a seperate PR

SeanKennyNF (approved):
For sure! Yeah the filter colours do look nice now :)

PR #183, opened 2024-02-18
Disabled JS scrolling on the consolidated dashboard + analyze page

### LINKED ISSUE #178: Disable the JS Scrolling on the consolidated dashboard + analyze page
Let's disable the JS Scrolling on the consolidated dashboard + analyze page and rely solely on the CSS scrolling. The css scrolling comes from these css classes on the divs `snap-start snap-always scroll-smooth` so as long as those are still there you're good.

- The big component that enables the JS scrolling is called `ScrollSectionGroup` and lives in `components/ui/scroll-section-group/scroll-section-group.tsx`. This is the component that has to go.
- Instead of using url search params to determine which section to scroll to, let's use the hash property to link to sections with certain ids in the html ([this is new to me but this looks like the right property](https://developer.mozilla.org/en-US/docs/Web/API/URL/hash))
- Basically if you just take every section in `redesignedArbovirusPageSections` and move them to where the `renderScrollSectionGroup` call is, I feel like all you'd have to do after that is clean up unused code and add ids to the sections in the html.

SeanKennyNF:
Resolves #178. Disabled JS scrolling on the consolidated dashboard + analyze page

Peek 2024-02-18 12-29


app/pathogen/arbovirus/wip/page.tsx
```diff
@@ -1,45 +1,46 @@
...
-  
+
   return (
     <div className="col-span-12 row-span-2 grid gap-0 grid-cols-12 grid-rows-2 grid-flow-col w-screen overflow-hidden border-box -my-4 -ml-4">
       <div className="overflow-y-scroll col-span-2 h-full row-span-2 border-black border-r-2">
         <Filters
           className="p-4"
-          excludedFields={
-            idOfCurrentlyViewedSection === RedesignedArbovirusPageSectionId.MAP ? [FilterableField.pathogen] : [] 
-          }
```
SeanKennyNF, line 39: So an interesting consequence of what I've done is that the pathogen filter will always be shown since I'm no longer tracking which section is in view. If removing the pathogen filter when you're looking at the map is a feature we're interested in having on the new dashboard, let's make a separate ticket for that because it's a bit more involved now.

app/pathogen/arbovirus/wip/sections.tsx
```diff
@@ -1,76 +1,5 @@
-import { RefObject } from "react";
-import { ArbovirusMap } from "../dashboard/(map)/ArbovirusMap";
-import { ArboDataTable } from "@/app/pathogen/arbovirus/analyze/ArboDataTable";
-import { VisualizationId, addToVisualizationInformation } from "../visualizations/visualizations";
-import { ZoomIn } from "lucide-react";
-import { useRouter } from "next/navigation";
-import { VisualizationsSection } from "./visualizations-section";
-import { RenderScrollSectionContentFunction } from "@/components/ui/scroll-section-group/scroll-section-group";
-
 export enum RedesignedArbovirusPageSectionId {
```
SeanKennyNF, line 1: Can't put this in the same file as the page because nextjs is expecting only one thing exported from a page. If I could do that I would delete this tiny file but I can't so here it is lol

app/pathogen/arbovirus/wip/page.tsx
HimanshuRanka: Sounds good, feel free to go ahead and make a ticket for that

app/pathogen/arbovirus/wip/page.tsx
```diff
@@ -1,45 +1,46 @@
...
         />
       </div>
-      {renderScrollSectionGroup()}
+      <div
+        className={"overflow-y-scroll snap-y scroll-smooth col-span-10 row-span-2"}
+      >
+        <section
+          id={RedesignedArbovirusPageSectionId.MAP}
```
HimanshuRanka, line 26: Note to update the name of this variable when we do make the big merge

app/pathogen/arbovirus/wip/sections.tsx
HimanshuRanka: That is ok, when we have more pages we will want our sections to exist all in one place so we can control it better. This will allow us to create objects that have deeper nesting and scrollable links if needed in the future. 

I will say maybe it would be better for this file to live elsewhere in the application (utils or constants)

Additionally could change it to ts file (not sure the impact of this but I think it is good practice)

utils/referrer-link-util.ts
```diff
@@ -1,4 +1,4 @@
-const safeReferrerLinkRegex = /^(\/[a-zA-Z0-9?=]+)+$/
+const safeReferrerLinkRegex = /^(\/[a-zA-Z0-9?=#]+)+$/
```
HimanshuRanka, line 1: +1 love this

HimanshuRanka (approved):
Some really nit changes that need to go in but lgtm otherwise (can do later also)

app/pathogen/arbovirus/wip/sections.tsx
SeanKennyNF: I think a `constants.ts` is a pretty reasonable approach so that's what I've chosen to send this to. It's at the base of the `app` folder which feels natural to me but if we want it moved somewhere else just let me know and I'll move it hahaha

Also yeah no impact on making it a `.ts` file, that sounds good to me and I did it as a part of this commit.

https://github.com/serotracker/dashboard/pull/183/commits/db5dcdd1fbed7f7006e387e6b78668e87d79a2a7

app/pathogen/arbovirus/wip/page.tsx
SeanKennyNF: Oh wait your PR wasn't switching over. Nevermind then, whichever one of us takes it can make sure this gets changed :)

HimanshuRanka (approved):

PR #184, opened 2024-02-18
Made the filter bar visible on fullscreen visualizations

### LINKED ISSUE #170: Make the filter bar visible when the visualization is expanded in the new UI
Currently  when u expand a visualization you cannot use the filters to fine tune it to what u may need. That is something I believe would be very valuable especially when we allow users to donwload our visualizations as images.

SeanKennyNF:
Resolves #170. Made the filter bar visible on fullscreen visualizations.

Peek 2024-02-18 13-10


app/pathogen/arbovirus/visualizations/visualizations.tsx
```diff
@@ -54,7 +54,6 @@ export interface VisualizationInformation {
   urlParameter: VisualizationUrlParameter;
   displayName: VisualizationDisplayName;
   renderVisualization: () => React.ReactNode;
-  classNameWhenFullscreen: string
```
SeanKennyNF, line 57: Yeah so when I added the filter bar I took another look at the classes and figured out why this extra padding was needed and fixed it so we don't need this anymore :sparkles:

app/pathogen/arbovirus/visualizations/page.tsx
```diff
@@ -23,26 +24,29 @@ export default function VisualizationsPage() {
...
-            }
-          }}
-          aria-label="Close Visualization"
-        >
-          <X />
-        </button>
-      </div>
-      <div className={cn("flex-1", visualizationInformation.classNameWhenFullscreen)}>
-        {visualizationInformation.renderVisualization()}
+    <div className="w-screen overflow-y-hidden grid grid-cols-12 grid-rows-2 -my-4 -ml-4 h-full-screen">
```
SeanKennyNF, line 27: Same problem as the `wip` page. I have to fight the layout of the pages we're going to remove. The negative margin can be removed when we make the `wip` page the main dashboard.

app/pathogen/arbovirus/visualizations/visualizations.tsx
HimanshuRanka: Curious: what was the reaosn?

HimanshuRanka (approved):

PR #185, opened 2024-02-19
Fixed the yellow fever chart in the change in median seroprevalence visualization

### LINKED ISSUE #171: Fix the Yellow Fever portion of the graph for the "Change in median seroprevalence" visualization
[Flagged by Harriet in the following thread.](https://3.basecamp.com/5061858/buckets/30589330/messages/6958342546#__recording_7046030678) There is some kind of browser specific issue that causes the Yellow Fever bars of the "Change in median seroprevalence" visualization to not display its bars on the Google Chrome Browser. Let's fix this issue and get the graph displaying nicely on both Google Chrome and Firefox.

Here are some images of the "Change in median seroprevalence" visualization captured on Firefox 122.0. This shows the graph displaying correctly.

image
image(1)

Here are some images of the "Change in median seroprevalence" visualization captured on Google Chrome 121.0. Note the Yellow Fever graph does not display its bars.

image
image

SeanKennyNF:
Resolves #171. Fixed the yellow fever chart in the change in median seroprevalence visualization

Peek 2024-02-18 17-43

### DIFF
--- a/app/pathogen/arbovirus/analyze/recharts/change-in-median-seroprevalence-over-time-graph.tsx
+++ b/app/pathogen/arbovirus/analyze/recharts/change-in-median-seroprevalence-over-time-graph.tsx
@@ -115,16 +115,18 @@ export const ChangeInMedianSeroprevalenceOverTimeGraph = (): React.ReactNode => 
       ([arbovirus, dataPointsForAParticularArbovirus]) => [
         arbovirus,
         groupDataPointsIntoTimeBuckets({
-          dataPoints: dataPointsForAParticularArbovirus.map((dataPoint) => ({
-            ...dataPoint,
-            groupingTimeInterval: {
-              intervalStartDate: parseISO(dataPoint.sampleStartDate),
-              intervalEndDate: parseISO(dataPoint.sampleEndDate),
+          dataPoints: dataPointsForAParticularArbovirus
+            .filter((dataPoint) => dataPoint.sampleStartDate && dataPoint.sampleEndDate)
+            .map((dataPoint) => ({
+              ...dataPoint,
+              groupingTimeInterval: {
+                intervalStartDate: parseISO(dataPoint.sampleStartDate),
+                intervalEndDate: parseISO(dataPoint.sampleEndDate),
+              },
+            })),
+            groupingBucketSize: {
+              years: 5,
             },
-          })),
-          groupingBucketSize: {
-            years: 5,
-          },
         }).groupedDataPoints,
       ]
     )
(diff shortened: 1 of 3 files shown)


app/pathogen/arbovirus/analyze/recharts/change-in-median-seroprevalence-over-time-graph.tsx
```diff
@@ -115,16 +115,18 @@ export const ChangeInMedianSeroprevalenceOverTimeGraph = (): React.ReactNode =>
       ([arbovirus, dataPointsForAParticularArbovirus]) => [
         arbovirus,
         groupDataPointsIntoTimeBuckets({
-          dataPoints: dataPointsForAParticularArbovirus.map((dataPoint) => ({
-            ...dataPoint,
-            groupingTimeInterval: {
-              intervalStartDate: parseISO(dataPoint.sampleStartDate),
-              intervalEndDate: parseISO(dataPoint.sampleEndDate),
+          dataPoints: dataPointsForAParticularArbovirus
+            .filter((dataPoint) => dataPoint.sampleStartDate && dataPoint.sampleEndDate)
```
SeanKennyNF, line 119: This was the problem :/

If you're wondering why Chrome was affected and Firefox was not, I think it's just some browser jank and I don't want to read into the behaviour of the different functions when you pass in undefined just to find out why this happened, I don't think that's a super good use of time.

I could make some kind of rule for when the start / end date isn't present but I actually read the study in question and there is a sampling end date in the text, we just haven't recorded one. Also the only reasonable solution here would be to set one to the other and I don't like that we have inconsistent strategies everywhere and throwing magic in different places for dealing with this problem.

My plan here instead is to read through as much as I can and try to get sampling start and end dates on everything and bring them to data. After that, I'll set up the ETL to reject anything with a missing sampling start date and end date and maybe generate some kind of report that can be sent to the data folks. Anyways, this is all dependent on how data feels but just dropping my feelings here.

package.json
```diff
@@ -43,7 +43,7 @@
     "react-day-picker": "^8.9.1",
     "react-dom": "18.2.0",
     "react-map-gl": "^7.1.6",
-    "recharts": "^2.10.1",
+    "recharts": "^2.12.0",
```
SeanKennyNF, line 46: Not necessary for the fix, just a good idea to keep this up to date.

app/pathogen/arbovirus/analyze/recharts/change-in-median-seroprevalence-over-time-graph.tsx
HimanshuRanka: Ok I agree! So far devs did not really touch the data. There is a field in the airtable that makrs a record as ready for extraction or not based on what data is required vs optional. I like the approach of report and not allowing null fields BUT we may need to do that if the Data team feels like we should

@harrietware any input?

HimanshuRanka (approved):

PR #187, opened 2024-02-24
Visualizations can be downloaded as images

### LINKED ISSUE #119: Make it so that visualizations can be downloaded as images
Implement functionality to convert visualizations to images. How?

SeanKennyNF:
Resolves #119

Peek 2024-02-24 16-05

Some sample images downloaded from the visualizations for you :)

estimate-count-by-who-region-and-arbovirus(1)

estimate-count-by-who-region-and-arbovirus

estimate-count-by-arbovirus-and-antibody-type

cumulative-estimate-count-over-time-by-sample-frame


app/pathogen/arbovirus/analyze/recharts/recharts-visualization.tsx
SeanKennyNF, line 1: Alright, real quick "what am I looking at" summary for you.

I consolidated the fullscreen visualizations and the visualizations on the WIP screen into one component and made the buttons and stuff dynamic because I don't want to drop the download code all over the place and it's really annoying to have different versions of the visualizations floating around everywhere. It's also a pain to maintain a whole bunch of different visualization headers. That's what all of this button config nonsense that I wrote is doing, it's for dynamic buttons so I can reuse the component. I didn't put the download button on the ones in the `/dashboard` and `/analyze` route because there's some frustrating css there and it just seems not worth the effort given that we're launching the WIP page really soon.

That's literally all `RechartsVisualization` and `VisualizationHeader` are. It's just consolidating some common stuff and making it consistent.

app/pathogen/arbovirus/analyze/recharts/use-download-visualization.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const useDownloadVisualization = (input: UseDownloadVisualizationInput): UseDownloadVisualizationOutput => {
+  const ref = useRef<HTMLDivElement>(null);
+
+  const downloadVisualization = (functionInput: DownloadVisualizationInput) => {
+    if(!ref.current) {
+      throw Error("Unable to generate image for visualization.")
+    }
+
+    domtoimage
```
SeanKennyNF, line 27: This library turned out to have almost instant download times while `htmltocanvas` was taking like 4-5 seconds. `htmltocanvas` had a bunch of messy interactions with recharts and `recharts-to-png` assumes that the element that `ref` is attached to is the recharts graph so I would lose the header and any padding around the graph which I can't do.

Anyways, that's what you're looking at here. This was the best solution for this by far.

app/pathogen/arbovirus/analyze/recharts/use-download-visualization.tsx
```diff
@@ -0,0 +1,47 @@
...
+      .toBlob(ref.current, {
+        bgcolor: "#FFFFFF",
+        filter: (input) => {
+          if(functionInput.elementIdsToIgnore.includes((input as HTMLElement)['id'])) {
```
SeanKennyNF, line 31: Obvious later but this is to make the image download without the icons.

app/pathogen/arbovirus/analyze/recharts/visualization-header.tsx
```diff
@@ -0,0 +1,134 @@
...
+  router: AppRouterInstance;
+}
+
+interface AllButtonConfigurations {
+  zoomInButton: ButtonConfig<ZoomInButtonAdditionalButtonConfig> & { id: string };
+  downloadButton: ButtonConfig<DownloadButtonAdditionalButtonConfig> & { id: string };
+  closeButton: ButtonConfig<CloseButtonAdditionalButtonConfig> & { id: string };
+}
+
+const DownloadButton = (props: DownloadButtonProps) => (
```
SeanKennyNF, line 49: There's probably a case to be made that these buttons and their types should live in separate files. I think that would be interesting so let me know if you'd like to see that!

app/pathogen/arbovirus/wip/visualizations-section.tsx
```diff
@@ -29,19 +29,24 @@ export const VisualizationsSection = () => {
...
+          key={visualizationInformation.id}
+          visualizationInformation={visualizationInformation}
+          className={cn(visualizationInformation.className, 'pb-14')}
+          buttonConfig={{
+            downloadButton: {
+              enabled: true,
+            },
+            zoomInButton: {
+              enabled: true,
+              referrerRoute: `/pathogen/arbovirus/wip#${RedesignedArbovirusPageSectionId.VISUALIZATIONS}`
```
SeanKennyNF, line 43: This is why I can't just read the route from NextJS. I want going back to send you to this particular section of WIP. If this isn't working for you it's because the CSS scrolling stuff needs to get merged for this to work but I still think it's a good idea to ship this and then when CSS scrolling stuff is shipped this will start working if that makes sense?

app/pathogen/arbovirus/wip/visualizations-section.tsx
```diff
@@ -29,19 +29,24 @@ export const VisualizationsSection = () => {
...
-              <ZoomIn />
-            </button>
-          </div>
-          {visualization.renderVisualization()}
-        </div>
+      return visualizationList.map((visualizationInformation) => (
+        <RechartsVisualization
+          key={visualizationInformation.id}
+          visualizationInformation={visualizationInformation}
+          className={cn(visualizationInformation.className, 'pb-14')}
```
SeanKennyNF, line 36: Padding necessary for correct dimensions on downloaded image and to prevent the graphs from overlapping with one another on the website. This used to be margin and that worked fine but I need it as padding because I want it to be part of the downloaded region if that makes sense.

app/pathogen/arbovirus/visualizations/page.tsx
```diff
@@ -26,28 +27,22 @@ export default function VisualizationsPage() {
...
-            aria-label="Close Visualization"
-          >
-            <X />
-          </button>
-        </div>
-        <div className="flex-1">
-          {visualizationInformation.renderVisualization()}
-        </div>
-      </div>
+      <RechartsVisualization
```
SeanKennyNF, line 30: On fullscreen Chrome appears to occasionally download it with the scroll bars showing. I think this is because the overflow is set on the graph and not some kind of parent component. Anyways, I'm going to make myself a ticket and come back to this. There's some other visualization stuff I want to hit then as well and this PR is big enough lol

HimanshuRanka (approved):
Looks good to me! Some of the impl details are hard to follow in a PR context but from a higher level seems like all good

HimanshuRanka (changes requested):
The close button still points to wip: 

image

HimanshuRanka:
Also, Can we make it so that the visualization section (the part of the page with all the visualizations is not limited to h-screen and instead just takes the size of all viz. That will avoid this from happening:

image

Over here we have the data page only partially scrolled but the visualization starts scrolling anyway.

app/about/faq/text.tsx
```diff
@@ -12,7 +12,7 @@ export enum FAQPageOptionId {
...
+    content: <p className='inline'> We collect all seroprevalence data from published peer reviewed research articles, preprints, reports, and media (unpublished grey literature). The source of any individual estimate can be viewed by clicking on the estimate on the map or by referencing the Source column in the table in the dashboard or the url column in a csv downloaded from the table. <Link className="inline text-link underline" target="_blank" href="/pathogen/arbovirus/dashboard#table">This is a link to the ArboTracker map</Link> and <Link className="inline text-link underline" target="_blank" href="/pathogen/arbovirus/dashboard#table">this is a link to the ArboTracker data table</Link>.</p>
```
SeanKennyNF, line 15: > The close button still points to wip:

Good catch, thank you sir. Resolved in d5583ae01fbb58e029dcf137778b46f07fe4c29b

SeanKennyNF, line 15: > Also, Can we make it so that the visualization section (the part of the page with all the visualizations is not limited to h-screen and instead just takes the size of all viz. That will avoid this from happening:

Yes we can! Resolved in cf40742b1ddf5d71ac66b5160e1f32fc4257235f!

HimanshuRanka (dismissed):

HimanshuRanka (approved):

PR #190, opened 2024-02-25
Improved the bucketing for change in median seroprevalence graph to make it attempt to consistently generate 10 bars.

### LINKED ISSUE #174: Make the number of buckets in the "Change in median seroprevalence over time" fixed to 10.
Currently the "Change in median seroprevalence over time" graph makes buckets of a fixed size of 5 years.

image

The graph looks fine when no time range is specified in the filter but if you filter on a small time range, the graph will come out looking like this:

image

What should probably happen instead is that filtering on a time range where there is a very small time range with studies should decrease the bucket size from 5 years to something smaller so that you can see a more granular breakdown and the graph becomes helpful even at smaller time ranges.

The plan for this would be that instead of having buckets of 5 years, you would set up the buckets so that there is a certain number of them. We're going with 10 for this ticket. If you don't have a filter applied, the bucket size might be 5 years but if you have a filter applied that makes the time range small, the bucket size might go down to 1 year.

It's worth noting that the bucket size would have to be different for each of the pathogens (because the time ranges encompassed by the studies are different per pathogen) so it would have to be on an individual graph level.

I think what makes this challenging is that the studies aren't continuous for most of the arboviruses so we can't use (end year - start year) / 10 because that would assume the data is continuous. If one of the years is missing we wouldn't show a bar for that year so if there were holes in the data you'd end up with less than 10 bars. This would mostly be relevant for arboviruses like Mayaro where we don't have a huge amount of studies or Zika which have decades with no studies in their graph. Here's an idea for an algorithm that I feel would get us the right number of bars most of the time.

- Generate a set of buckets for 1 year, 2 years, 3 years, 4 years, and 5 years whenever you're generating buckets for the graph.
- When you're putting the data points into the buckets, do it for every set of buckets not just the 5 year ones
- Filter out the empty buckets like we do today
- Compare the number of buckets for each granularity (1 year, 2 years, etc.) and find the set of buckets whose number of bars is as small as possible while still being over or equal to 10. If there is no granularity that will give you 10 or above buckets, just pick the 1 year bucket.
- Use that set of buckets for generating the graph for this pathogen

It would be interesting to see if we can make the graph go down to month level granularity if your time filter is small enough, but let's leave that for another ticket. For now, let's make the minimum bucket size 1 year and if there aren't 10 buckets even with one year, it just shows you the one year one.

Come talk to me if you think the code for the bucketing is confusing. I wrote the original bucketing code so you can send all angry letters about how unreadable it is to me as well :)

SeanKennyNF:
Resolves #174. Best shown with before and after pictures with the old version. The graph now attempts to generate as close to 10 bars as it can using the interval sizes available instead of always using a bucket size of 5 years. This is most apparent when there is a time filter applied as now the graph gets more granular as you use a smaller time filter which is pretty neat.

Old Version             |  New Version
:-------------------------:|:-------------------------:
image | image
image | image
image | image
image | image


app/pathogen/arbovirus/analyze/recharts/change-in-median-seroprevalence-over-time-graph.tsx
```diff
@@ -49,9 +52,13 @@ interface GenerateBucketsForTimeIntervalOutput {
...
   const buckets: TimeInterval[] = [];
-  
+
+  if(input.groupingBucketSize.years === 0) {
+    throw new Error("Unable to group data points with a bucket size of zero years.")
+  }
+
   while(isBefore(currentIntervalStartDate, input.latestDate)) {
-    const currentIntervalEndDate = endOfYear(addYears(currentIntervalStartDate, input.groupingBucketSize.years));
+    const currentIntervalEndDate = endOfYear(addYears(currentIntervalStartDate, input.groupingBucketSize.years - 1));
```
SeanKennyNF, line 61: Turns out I accidentally made a bucket size of 5 years equal 6 because of this algo. Whoops, fixed now. Only realized it when I tried to put 1 year into this lol

app/pathogen/arbovirus/analyze/recharts/change-in-median-seroprevalence-over-time-graph.tsx
```diff
@@ -86,16 +93,34 @@ const groupDataPointsIntoTimeBuckets = <
...
-  const { buckets } = generateBucketsForTimeInterval({
-    earliestDate,
-    latestDate,
-    groupingBucketSize: input.groupingBucketSize
-  })
+  if(input.validBucketSizes.length === 0) {
+    return { groupedDataPoints: [] };
+  }
+
+  const { groupedDataPoints } = input.validBucketSizes.map((bucketSize) => {
```
SeanKennyNF, line 100: This part is kind of confusing but the idea is that it generates buckets with 1 year, 2 years, 3 years, 4 years, and 5 years and then figures out how many bars each of them would have and then picks the bucket size that got closest to `desiredBucketCount`, in this case 10.

app/pathogen/arbovirus/analyze/recharts/change-in-median-seroprevalence-over-time-graph.tsx
```diff
@@ -137,7 +167,9 @@ export const ChangeInMedianSeroprevalenceOverTimeGraph = (): React.ReactNode =>
       <div className="h-full flex flex-row flex-wrap">
         {Object.keys(dataGroupedByArbovirusThenByTimeBucket).map((arbovirus, index) => {
           const dataForArbovirus = dataGroupedByArbovirusThenByTimeBucket[arbovirus].map(({interval, dataPoints}) => ({
-            intervalAsString: `${interval.intervalStartDate.getFullYear()}-${interval.intervalEndDate.getFullYear()}`,
+            intervalAsString: interval.intervalStartDate.getFullYear() !== interval.intervalEndDate.getFullYear() ?
+              `${interval.intervalStartDate.getFullYear()}-${interval.intervalEndDate.getFullYear()}` :
+              `${interval.intervalStartDate.getFullYear()}`,
```
SeanKennyNF, line 172: Never considered bucket size of 1 in the original code whoops hahaha

app/pathogen/arbovirus/analyze/recharts/change-in-median-seroprevalence-over-time-graph.tsx
```diff
@@ -124,9 +149,14 @@ export const ChangeInMedianSeroprevalenceOverTimeGraph = (): React.ReactNode =>
...
-              years: 5,
-            },
+            desiredBucketCount: 10,
+            validBucketSizes: [
+              { years: 5 },
+              { years: 4 },
+              { years: 3 },
+              { years: 2 },
+              { years: 1 }
+            ]
```
SeanKennyNF, line 159: Each bucket size has a certain cost because it has to go through the entire array of studies for each bucket size so I'm not using anything above 5 because I found that in practice anything above 5 is pretty rare to get picked.

Reading this again, I think this algorithm has room a lot of room for optimization so I'm going to come back later and see what can be done. It's `O(number of studies * number of total years spanned for studies * number of possible bucket sizes)` right now for those who are curious but I think I can do a lot better.

HimanshuRanka (approved):
LGTM 🚀

Issue #106, opened 2023-12-18
Add a filter for `Serotype` on the dashboard


SeanKennyNF:
> > Feedback: Dengue by subtype, to layer infection with diff serotypes, as a possible filter

> Harriet: We've added this variable in airtable, need to add to the dashboard

Blocked by https://github.com/PathoTracker/Pathotracker/issues/105

Let's add a filter by serotype. Serotype is a list and many non-dengue studies don't have serotypes listed. I think the way the filter would work is that it would work like the `pathogen` filter where every possible value is selected at the start and then you can choose to remove some of them. Studies would stay as a part of the filter if at least one of the serotypes in their list is included in the list in the filter.

If the study doesn't have a `serotype` (most of the studies) the serotype filter shouldn't affect it at all in my opinion.

It would be neat to (and probably easy?) add this to the analysis table as well I think. We could probably copy our code for `Antibodies` and also have the pretty colours :)

harrietware:
This all sounds great! Serotype is dengue-specific so only dengue studies will have serotypes listed :)

EliasPinno:
look in `useArboFilters` and `useArboData` to see how it works. Dm @SeanKennyNF if you are lost.

PR #192, opened 2024-02-26
Added a filter for pediatric age group.

### LINKED ISSUE #148: Add a filter on `pediatricAgeGroup` on the front end
Blocked by https://github.com/PathoTracker/Pathotracker/issues/147

Let's add a filter on the `pediatricAgeGroup` field. This filter should only appear if the only selected age group filter is age 0-17.

SeanKennyNF:
Added a filter for pediatric age group! Resolves #148 
Check out the comments in the PR, there's a couple asterisks here but I still think it's neat hahaha

Peek 2024-02-25 17-33


app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -171,11 +172,20 @@ interface FiltersProps {
...
 export function Filters(props: FiltersProps) {
+  const { selectedFilters } = useContext(ArboContext);
+
   const excludedFields = props.excludedFields ?? [];
 
+  const selectedAgeGroups = selectedFilters['ageGroup'] ?? [];
+
+  if(!(selectedAgeGroups.length === 1 && selectedAgeGroups[0] === 'Children and Youth (0-17 years)')) {
+    excludedFields.push(FilterableField.pediatricAgeGroup);
+  }
```
SeanKennyNF, line 183: 100% agree not ideal to be hardcoding this stuff in the front end because if it changes in Airtable it will break the pediatric age group filter. I definitely don't think this is the last iteration of this, I just want to give us something to iterate on lol

SeanKennyNF: An interesting consequence of the way our filter system works is that I can't say "filter on age group (0-4) and age group (18-64)" because that would involve adding the pediatric age group filter and it would filter out all the studies with a pediatric age group of null which would be all of the (18-64) ones.

Thought that was kind of interesting. I'm not implementing a fix in this PR because I think this PR is big enough as-is but thought I would mention it for an interesting feature we might want to consider later.

components/ui/data-table/use-data-table-styles.tsx
```diff
@@ -18,7 +18,7 @@ export const useDataTableStyles = <TData, TValue>(input: useDataStylesInput<TDat
   const generateClassnameForHeader = (input: GenerateClassnameForHeaderInput) => {
     const { columnId } = input;
 
-    const baseClassnameForHeader = "border-b bg-white";
+    const baseClassnameForHeader = "border-b bg-white whitespace-nowrap";
```
SeanKennyNF, line 21: The reason I want this is because "Pediatric Age Group" is a long word and if you don't do this it will display on three lines in the header of the table and it looks really not good. This makes it all fit on one line and expands the header horizontally if it needs more space.
image

contexts/arbo-context/arbo-context.tsx
```diff
@@ -0,0 +1,134 @@
+"use client";
```
SeanKennyNF, line 1: Did a bit of a refactor of this file. I've moved any actions that have to be taken after the filters update to files for steps because this file was getting difficult to work with, I had some major code cleanup to do with the country bounding box code, and the order of operations was starting to matter so I wanted to make a file that makes the order of operations incredibly clear. Anyways, new function is called on line 47 of this file. Every other change in this file is removals.

contexts/arbo-context/arbo-context.tsx
```diff
@@ -0,0 +1,134 @@
...
+  action: ArboAction,
+  map: MapRef | undefined
+) => {
+  switch (action.type) {
+    case ArboActionType.UPDATE_FILTER:
+      return handleArboFilterUpdate({
+        state,
+        action,
+        map
+      }).state;
```
SeanKennyNF, line 51: New function. Everything else is the same or was removed.

contexts/arbo-context/arbo-filter-update-steps.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export interface HandleArboFilterUpdateOutput {
+  state: ArboStateType,
+  action: ArboAction,
+  map: MapRef | undefined
+}
+
+export const handleArboFilterUpdate = (input: HandleArboFilterUpdateInput): HandleArboFilterUpdateOutput => {
+  return pipe(
```
SeanKennyNF, line 22: https://gcanti.github.io/fp-ts/modules/function.ts.html#pipe
Equivalent to this:
```
applyNewSelectedFiltersStep(
  updatePediatricAgeGroupFilter(
    adjustMapPositionStep(
      addActionToSelectedFiltersStep(
        input
      )
    )
  )
)
```
Just reads better if you have a lot of functions and lists the functions in the order they are called not reverse the order they are called which I'm a big fan of.

This is what the ETL uses too and I really love this function.

contexts/arbo-context/filter-update-steps/add-action-to-selected-filters-step.tsx
```diff
@@ -0,0 +1,18 @@
+import { HandleArboFilterUpdateInput, HandleArboFilterUpdateOutput } from "../arbo-filter-update-steps";
+
+export const addActionToSelectedFiltersStep = (
```
SeanKennyNF, line 3: Doing this step first isn't the old way of doing it but it makes the other steps simpler (see `adjustMapPositionStep`)

contexts/arbo-context/filter-update-steps/adjust-map-position-step.tsx
SeanKennyNF, line 1: The old version of this was a lot grosser and would look at the payload's filter as well as the selected filters and combine. This is simpler, only looks at selected filters because the order of steps has changed. I also like this stuff living in it's own file because WHO region stuff has to go here too now and I don't want to clutter up the arbo context.

contexts/arbo-context/filter-update-steps/apply-new-selected-filters-step.tsx
SeanKennyNF, line 1: Straight copy-paste.

contexts/arbo-context/filter-update-steps/update-pediatric-age-group-filter.tsx
```diff
@@ -0,0 +1,29 @@
+import {
+  HandleArboFilterUpdateInput,
+  HandleArboFilterUpdateOutput,
+} from "../arbo-filter-update-steps";
+
+export const updatePediatricAgeGroupFilter = (
```
SeanKennyNF, line 6: The reason why I need this is because if you deselect 'Children and Youth (0-17 years)' or select another age group as well, I have to get rid of the pediatric age group filter because even though the filter dissapears from the UI I need to remove it from our logic as well so that you don't just see no data or deselect the "Children and Youth" and be confused as to why you're still seeing only seeing data for 0-4 year olds.

app/pathogen/arbovirus/dashboard/filters.tsx
SeanKennyNF: I bet #106 would need to basically work the same way as this except with DENV/pathogen instead of 0-17/Age group so that's maybe an interesting time to copy-paste this code and maybe see if we can make something more generalized once there's two copies of the code? Just kind of thinking out loud lol

SeanKennyNF: When we're doing SarsCoV2 stuff, I'm going to see if there's maybe some kind of thing we can do in the backend for things like this? I'd have to give it some more thought hahaha

HimanshuRanka: All sounds good to me, and yeah for now it is a-ok. I like the idea of coming back to it when we do the DENV 1 2 3 4 stuff

contexts/arbo-context/arbo-filter-update-steps.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { addActionToSelectedFiltersStep } from "./filter-update-steps/add-action-to-selected-filters-step";
+import { applyNewSelectedFiltersStep } from "./filter-update-steps/apply-new-selected-filters-step";
+
+export interface HandleArboFilterUpdateInput {
+  state: ArboStateType,
+  action: ArboAction,
+  map: MapRef | undefined
+}
+
+export interface HandleArboFilterUpdateOutput {
```
HimanshuRanka, line 15: what is the difference between this type and the type above?

contexts/arbo-context/arbo-filter-update-steps.tsx
HimanshuRanka: +1 Yeah this is super cool! I did not know this existed

contexts/arbo-context/arbo-filter-update-steps.tsx
```diff
@@ -0,0 +1,29 @@
...
+  state: ArboStateType,
+  action: ArboAction,
+  map: MapRef | undefined
+}
+
+export const handleArboFilterUpdate = (input: HandleArboFilterUpdateInput): HandleArboFilterUpdateOutput => {
+  return pipe(
+    input,
+    addActionToSelectedFiltersStep,
+    adjustMapPositionStep,
```
HimanshuRanka, line 25: nit: Only note would be to remove the step from the function names. I think using pipe indicates they are steps and the names are descriptive enough without the step. Could be good incase we want to do stuff outside the pipe (idk if needed)

HimanshuRanka (approved):
LGTM, just left a couple small notes

contexts/arbo-context/arbo-filter-update-steps.tsx
SeanKennyNF: Yeah you're right, repeating the type like this doesn't really make sense. I'm going to keep two distinct types just in case they are ever different but I'm setting one equal to the other :+1:

ad19f0241cd080a899e982f134caac34b89f93bd

contexts/arbo-context/arbo-filter-update-steps.tsx
SeanKennyNF: Yeah I think that's fair, I just realized I was being inconsistent with step/nostep anyhow so let's be consistent hahaha. Changed the name of all of the functions and file names for the individual steps. Kept the name of the folder as `filter-update-steps` and `arbo-filter-update-steps` but let me know if you think changing those makes sense too :)

ca80b92c183331b97ef5f68cacec17258c18c48d

HimanshuRanka (approved):
🚀

PR #195, opened 2024-03-02
Migrating the `about/the-team` page to use the auto generated types.

EliasPinno:
Example PR: this was nice and simple to do correctly! Requires the original auto generated types PR to merge first.


gql/graphql.ts
```diff
@@ -109,12 +109,12 @@ export type ArbovirusFilterOptionsQueryVariables = Exact<{ [key: string]: never;
...
 
-export type GroupedTeamMembersQueryQueryVariables = Exact<{ [key: string]: never; }>;
+export type GroupedTeamMembersQueryVariables = Exact<{ [key: string]: never; }>;
 
 
-export type GroupedTeamMembersQueryQuery = { __typename?: 'Query', groupedTeamMembers: Array<{ __typename?: 'TeamMemberGroup', label: string, teamMembers: Array<{ __typename?: 'TeamMember', firstName: string, lastName: string, email?: string | null, linkedinUrl?: string | null, twitterUrl?: string | null, affiliations: Array<{ __typename?: 'Affiliation', label: string }> }> }> };
+export type GroupedTeamMembersQuery = { __typename?: 'Query', groupedTeamMembers: Array<{ __typename?: 'TeamMemberGroup', label: string, teamMembers: Array<{ __typename?: 'TeamMember', firstName: string, lastName: string, email?: string | null, linkedinUrl?: string | null, twitterUrl?: string | null, affiliations: Array<{ __typename?: 'Affiliation', label: string }> }> }> };
```
EliasPinno, line 115: I changed the name of the query to avoid having the QueryQuery type.

app/about/the-team/page.tsx
```diff
@@ -7,22 +8,6 @@ import { Linkedin, Mail, Twitter } from "lucide-react";
...
-  linkedinUrl: string | null | undefined;
-  twitterUrl: string | null | undefined;
-  affiliations: Array<{
-    label: string;
-  }>
-}
-
 interface TeamMemberInfoCardProps {
   teamMember: TeamMember;
 }
```
EliasPinno, line 13: Just changed the import here: very simple.

app/about/the-team/page.tsx
```diff
@@ -7,22 +8,6 @@ import { Linkedin, Mail, Twitter } from "lucide-react";
...
-  firstName: string;
-  lastName: string;
-  email: string | null | undefined;
-  linkedinUrl: string | null | undefined;
-  twitterUrl: string | null | undefined;
-  affiliations: Array<{
-    label: string;
-  }>
-}
-
```
EliasPinno, line 25: Deleted these types: now we read from the front end.

app/about/the-team/page.tsx
```diff
@@ -1,4 +1,5 @@
 "use client";
+import { TeamMember, TeamMemberGroup } from "@/gql/graphql";
```
EliasPinno, line 2: Where we import the types from!

hooks/useGroupedTeamMemberData.tsx
```diff
@@ -1,9 +1,10 @@
 import { useQuery } from "@tanstack/react-query";
 import { gql } from "@apollo/client";
 import { request } from 'graphql-request';
+import { GroupedTeamMembersQuery } from "@/gql/graphql";
 
-export const groupedTeamMembersQuery = gql`
-  query groupedTeamMembersQuery {
+export const groupedTeamMembers = gql`
+  query groupedTeamMembers {
```
EliasPinno, line 7: Query name changed.

HimanshuRanka:
Would love a quick recap/knowledge transfer on your work on these 2 PR's tomorrow!

app/about/the-team/page.tsx
SeanKennyNF: My one word of caution here would be that you're importing the type that is derived from the **graphql schema**, not the type that's derived from the **query**. This means that if we add a new field to `TeamMember` on the backend but don't request it on the front end, the build will break because `groupedTeamMembers['teamMembers']` would no longer be assignable to `TeamMember[]` since that type has more fields now. The only reason this typecast works is because this query is requesting every single field on `TeamMember`, if that stopped being true, this would stop working.

Anyways, I'm not saying this is something we need to fix right now but if you want to fix this your two options are:
- Define a new `TeamMember` type that is `type TeamMember = groupedTeamMembersQuery['teamMembers'][number]` (or something like that). This will be based on your query type and we'll be a-ok if you add new fields on the backend
- Redefine the `TeamMember` type including only fields that are used by this component.

Both work, this also works for now just a quick warning hahahaha

SeanKennyNF (approved):

SeanKennyNF (approved):
:fire:

PR #197, opened 2024-03-02
Issue 189 landing page wire frame

HimanshuRanka:


app/page.tsx
```diff
@@ -1,50 +1,190 @@
+import SeroMap from "../public/SeroMap.png";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import {
+  faMosquito,
+  faVirus,
+  IconDefinition,
+} from "@fortawesome/free-solid-svg-icons";
```
HimanshuRanka, line 7: I also think font awesome has a much more robust collection of icons, so maybe we should switch to that one completely instead of lucide

components/customs/arboStats.tsx
```diff
@@ -0,0 +1,15 @@
+"use client"
+
+import { useArboDataStatistics } from "@/hooks/useArboStatistics";
+
+function ArboStats() {
+    const arboDataStats = useArboDataStatistics();
```
HimanshuRanka, line 6: Big debate: should I instead be prefetching, dehydrating and rehydrating the data here? I technically do it one level up where this component is called, but do I need to do it here instead and convert this to a server component? Hmmm

contexts/custom-query-client-provider.tsx
```diff
@@ -0,0 +1,25 @@
+'use client'
+
+import getQueryClient from "@/components/customs/getQueryClient"
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+
+
+function CustomQueryClientProvider({ children }: { children: React.ReactNode }) {
+    // Why cant we use getQueryClient here?
+    const client = new QueryClient({
```
HimanshuRanka, line 9: we have a getQueryClient utility function. However I noticed in all our contexts we use this other query client that we generate each time --> This is not the best I think we may want a hook that creates and supplies a query client to our application.

I assume the issue with getQueryClient is that it lives on the server and is the query client for the server side data fetching. We need to init a new one on the client side so our client also has a query client and then use that (useQueryClient but it has the same functionality as getQueryClient?)

Sometime this server vs client stuff really throws me for a loop because there is not a very super clear distinction other than certain files being tagged as 'use_client' ... makes me wonder

app/page.tsx
```diff
@@ -1,50 +1,190 @@
...
+            <div className="flex flex-col justify-between w-full rounded-md p-4 mb-4 h-1/2 text-background bg-white/90">
+              <div className="w-full">
+                <h1 className=" w-fit p-2 rounded-md">SeroTracker</h1>
+                <h3 className="rounded-md p-2">
+                  Your Go to Source for COVID-19 and Arbovirus Seroprevalence
+                  Data
+                </h3>
+                <div className="p-2 rounded-md">
+                  <p className=" w-fit mb-2">
+                    We synthesize findings from thousands of COVID-19 and
```
HimanshuRanka, line 96: Copy is temporary placeholder, will be updated in the next week or so

contexts/theme-provider.tsx
```diff
@@ -0,0 +1,32 @@
...
+          document.documentElement.style.setProperty('--background', style.getPropertyValue('--sc2'));
+          document.documentElement.style.setProperty('--background-hover', style.getPropertyValue('--sc2-hover'));
+        } else {
+          document.documentElement.style.setProperty('--background', style.getPropertyValue('--sero'));
+          document.documentElement.style.setProperty('--background-hover', style.getPropertyValue('--sero-hover'));
+        }
+      }, [pathname]);
+    
+    return (
+        <div>
```
HimanshuRanka, line 26: Honestly not 100% sure if there is a better way of doing this but this works p well rn and I do not see any problems with it

SeanKennyNF: Yeah totally fine imo :)

SeanKennyNF (approved):
Cool! Couple suggestions + qs

components/customs/arboStats.tsx
```diff
@@ -0,0 +1,21 @@
...
+      <b className="px-1">{num?.toString().replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1,') ?? 0}</b>
+      );
+}
+
+function ArboStats() {
+    const arboDataStats = useArboDataStatistics();
+
+    return (
+        <h3 className="flex text-background bg-white rounded-md px-16 justify-center p-8 whitespace-nowrap">
+          We have data from {formatNumber(arboDataStats.data?.arbovirusDataStatistics.patricipantCount)} Participants accross {formatNumber(arboDataStats.data?.arbovirusDataStatistics.estimateCount)} Estimates from {formatNumber(arboDataStats.data?.arbovirusDataStatistics.sourceCount)} Sources spanning {formatNumber(arboDataStats.data?.arbovirusDataStatistics.countryCount)} countries and territories
```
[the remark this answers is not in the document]
HimanshuRanka: Or just a combined stat for both something like: "We have all this data across all our dashboards. Explore it further" 

Since the main purpose of this is to be like SO MUCH DATA WHAT A WOW COME USE OUR PLATFORM

contexts/theme-provider.tsx
```diff
@@ -0,0 +1,32 @@
...
+        {children}
+        </div>
+    )
+}
+
+export default ThemeProvider;
```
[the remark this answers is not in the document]
HimanshuRanka: This is actually a really really good point. I have not thought about that, I will go ahead and make the change

app/page.tsx
```diff
@@ -1,50 +1,190 @@
...
+            */}
+                {process.env.NEXT_PUBLIC_SARS_COV_2_TRACKER_ENABLED && (
+                  <TrackerButton
+                    titleSuffix="SC2"
+                    description="SeroTracker description"
+                    icon={faVirus}
+                    bgColor="bg-sc2virus"
+                    textColor="text-sc2virus"
+                    href={"/pathogen/sarscov2/dashboard"}
+                  />
```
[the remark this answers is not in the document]
HimanshuRanka: This is a good idea, and I will just label it legacy SC2 Tracker when the thing is disabled

components/customs/arboStats.tsx
HimanshuRanka: As far as I understand (and I am still navigating the waters of react query + SSR) what is happening here is the server calls the endpoint and gets the data. When loaded on the client, the client side cache is hydrated with the data that was fetched from the server (sent over as part of the defydratedState variable when site loads) and is hence loaded from cache

The only reason I wondered if I should do it here is the instance this component is accessed from another page we would have to hydrate it there seperately, but then again not sure. 

I think from this discussion I believe it best to dehydrate and rehydrate here to make the component truly reusable

HimanshuRanka: I am an idiot, we will never need this on the client side, I can just fetch on the server, populkate the value and send a static component to the client LOL

components/customs/arboStats.tsx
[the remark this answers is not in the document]
HimanshuRanka: I would prefer to make that change later as at the moment it really is specifically arboStats, when we change it in the future we can change the name! (just a personal preference so we know ++ maybe we want to show these stats on arbo/sc2treacker respectively and may want 2 seperate endpoints for it)

hooks/useArboStatistics.tsx
```diff
@@ -0,0 +1,23 @@
...
+        estimateCount,
+        countryCount
+    }
+  }
+`
+
+export const arbovirusDataStatisticsQueryKey = "arbovirusDataStatistics";
+
+export function useArboDataStatistics() {
+  return useQuery<any>({
```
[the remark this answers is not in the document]
HimanshuRanka: Gonna need a little guidance here. Do I need to run some type of codegen command to get these types, is that part of our build process... ?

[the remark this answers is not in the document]
HimanshuRanka: Done

hooks/useArboStatistics.tsx
```diff
@@ -0,0 +1,24 @@
...
+        estimateCount,
+        countryCount
+    }
+  }
+`
+
+export const arbovirusDataStatisticsQueryKey = "arbovirusDataStatistics";
+
+export function useArboDataStatistics() {
+  return useQuery<ArbovirusDataStatistics>({
```
SeanKennyNF, line 20:
```suggestion
  return useQuery<ArbovirusDataStatisticsQuery>({
```
Linking the comment I left on Elias' PR because the exact same logic applies here. https://github.com/serotracker/dashboard/pull/195#discussion_r1510023795

Basically I believe that `ArbovirusDataStatisticsQuery` is the correct type here and that comment I linked above explains why. It's a coincidence that the `ArbovirusDataStatisticsQuery` and the `ArbovirusDataStatistics` types match but as time goes on that might not be true so we're probably better off patching this up now.

SeanKennyNF (commented):
Something is wrong with mapbox in the preview for this PR. Do you know what's going on there?

image

EliasPinno (changes requested):
image

Chrome: Windows. 1920x1080 at 150% scale.

EliasPinno:
If I shrink my screen to make it thin:

image

Problem I think is just that the white box does not grow to the size of the text in it.

HimanshuRanka:
> Visit Preview

I agree with these issues mentioned (and am aware of them) but responsiveness across the webpage is something we have not at all considered and is probably something we want to put onto the radar.

My question to you there would be - would you still want me to address this here or should we do responsiveness as a whoel seperately?

EliasPinno (approved):

PR #198, opened 2024-03-03
Added zooming when a WHO region is filtered on.

### LINKED ISSUE #87: Filtering on a WHO region should zoom the map to that WHO region
Very similar to https://github.com/PathoTracker/Pathotracker/issues/18. I would check out https://github.com/PathoTracker/Pathotracker/pull/19 if you want to see what the code to pan the camera to a country looks like because this would be really similar. For the bounding box, you can use google maps to find around the lat and lng of certain places and kind of put the box together yourself by finding coordinates on their map.

SeanKennyNF:
Resolves #87.

Peek 2024-03-02 19-28


contexts/arbo-context/filter-update-steps/adjust-map-position.tsx
```diff
@@ -35,10 +37,20 @@ const getAllBoundingBoxesFromSelectedFilters = (
       (boundingBox: BoundingBox | undefined): boundingBox is BoundingBox =>
         !!boundingBox
     );
+  const selectedWHORegions = selectedFilters["whoRegion"] ?? [];
+  const boundingBoxesFromSelectedWHORegions = selectedWHORegions
+    .map((unRegion) =>
```
HimanshuRanka, line 42: ❓ This is throwing me for a loop ahaha, unRegion !== whoRegion right? so should this here be .map((whoRegion)) or am I missing something

HimanshuRanka (approved):
Not sure about the naming of the unRegion whoRegion thing but I think (after looking at the whole PR) that it should have been whoRegion.

contexts/arbo-context/filter-update-steps/adjust-map-position.tsx
SeanKennyNF: Yup you're 100% correct. Whoops, sorry skimmed this a bit too hard hahaha

f4ae019ee20c9627d84153ec594739b92f3300ab

HimanshuRanka (approved):
To the Mooooooon

EliasPinno (approved):

PR #199, opened 2024-03-08
135 design and implement a loading skeleton

EliasPinno:
Created a completed loading screen for the application.

Visual of how it looks (with an exaggerated loading time)

loading_skeleton_gif

The middle bar does not delete on the real website (that is gif video compression making me more angry then the sun)


HimanshuRanka:
For whatever reason this does not cover the case when you are navigating from arbo -> sc2. is that only on my end?

HimanshuRanka:
> For whatever reason this does not cover the case when you are navigating from arbo -> sc2. is that only on my end?

I think you need to put the skeleton one level down (and or? need to test if we need both or just 1 but yeah since arbo and sero share the same layount but are within the larger layout the loading fallback is not triggered (at pathogen lvel)

HimanshuRanka (changes requested):
Cross pathogen loading fallback needed

EliasPinno:
> For whatever reason this does not cover the case when you are navigating from arbo -> sc2. is that only on my end?

Navigating from: https://serotracker-dashboard-ifroojkhe-serotracker.vercel.app/pathogen/arbovirus/dashboard#MAP

To: 
https://serotracker-dashboard-ifroojkhe-serotracker.vercel.app/pathogen/sarscov2/dashboard#MAP

I do see the loading skeleton on my end. Using windows and chrome. 

You are right that while the map loads, I can't see the loading screen: but I think that is due to how the maps pins get populated? 

@SeanKennyNF what are you seeing?

app/loading.tsx
```diff
@@ -0,0 +1,13 @@
+import { Skeleton } from "@/components/ui/skeleton"
```
SeanKennyNF, line 1: While the skeleton looks good but I still think I have a poor understanding of when the skeleton will and won't appear and have got a couple questions.

- I don't see why cross-pathogen would be a problem, those pages actually don't share the same layout so I wouldn't expect any issues with those two pages together. I was unable to make this skeleton appear at all in my testing.
- I think you're right about the map pins Elias, stuff like loading times on the queries being made and parts of the map being rendered don't appear to be covered by a skeleton like this.
- This loading skeleton is at the base of the project. Does it also apply to the about page, the landing page, etc.?

HimanshuRanka:
I still do not see it going 

From https://serotracker-dashboard-git-135-design-and-imp-77315b-serotracker.vercel.app/pathogen/arbovirus/dashboard#MAP

To: https://serotracker-dashboard-git-135-design-and-imp-77315b-serotracker.vercel.app/pathogen/sarscov2/dashboard

HimanshuRanka:
> rendering

So using the links you sent me, when I change the link in the url it does show me the skeleton, but when using the header to navigate I do not see it

app/loading.tsx
HimanshuRanka: I think specifically because it is at the base, it only applies to the highest level of navigation. If we were to recreate the same loading.js and throw it into the pathogen dir then I think we will see the expected behaviour

components/ui/loading-page.tsx
```diff
@@ -0,0 +1,12 @@
+'use client'
```
HimanshuRanka, line 1: Why not just throw this back up in loading.tsx (is there a need for it to be a seperate componenet?)

EliasPinno:
So the loading thing works locally, or when navigating to the sarscov2 page. But on the vercel deployment, arbotracker loads too fast to actually see the loading page before components start loading. But we have loading screens when navigating into arbotracker.

SeanKennyNF:
Going ahead and cleaning up really old PRs. Honestly, the structure of the frontend has changed so much it would be very painful for this to get merged. If these are changes we want to make later, it's probably faster to restart :)

PR #200, opened 2024-03-09
Made the api key private

HimanshuRanka:
It was only being used in one place so easy peasy lmnsqzy

### DIFF
--- a/components/ui/pathogen-map/pathogen-map.tsx
+++ b/components/ui/pathogen-map/pathogen-map.tsx
@@ -142,7 +142,7 @@ export function PathogenMap<
       minZoom={2}
       maxZoom={14}
       interactiveLayerIds={[...layers.map((layer) => layer.id), 'country-highlight-layer']}
-      mapboxAccessToken={process.env.NEXT_PUBLIC_MAPBOX_API_KEY as string}
+      mapboxAccessToken={process.env.MAPBOX_API_KEY as string}
       onMouseEnter={onMouseEnter}
       onMouseDown={onMouseDown}
       onMouseLeave={onMouseLeave}


components/ui/pathogen-map/pathogen-map.tsx
```diff
@@ -142,7 +142,7 @@ export function PathogenMap<
       minZoom={2}
       maxZoom={14}
       interactiveLayerIds={[...layers.map((layer) => layer.id), 'country-highlight-layer']}
-      mapboxAccessToken={process.env.NEXT_PUBLIC_MAPBOX_API_KEY as string}
+      mapboxAccessToken={process.env.MAPBOX_API_KEY as string}
```
SeanKennyNF, line 145: Vercel preview not working for me :(

image

SeanKennyNF: We probably have to figure out something a bit more clever sadly :/

SeanKennyNF: Oh also just checked vercel, I would def make sure this is set in production too.

HimanshuRanka: yeah just saw that damn... ok will take a look

HimanshuRanka: I guess it wasnt ezpz lmnsqzy

HimanshuRanka:
This is not possible. Instead I had to just generate a public api key (this is safe to expose) and add url restrictions to it -- essentially from which urls it should accept requests etc. So just going to change the API key in vercel. The old api key still works and can be used on local anyway

PR #205, opened 2024-03-16
106 add a filter for serotype on the dashboard

EliasPinno:
Adds serotype (which was on backend) to front end.

I also changed the way the pediatric age group filter works, resolving #211 as well.

Also just a heads up on something a little bit tough to show: selecting a pediatric age group can give the ILLUSION of growing one of the pubbles (one goes from 92 to 108) but the total number of bubbles in the area shrinks, and what actually happens is that the second bubble steals from the first one because the first one is now smaller. Something to keep note of. I verified (with a good amount of certainty) that the filters are indeed actually reducing the number of records when used together.


app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -226,10 +227,16 @@ export function Filters(props: FiltersProps) {
...
-
-  if(!(selectedAgeGroups.length === 1 && selectedAgeGroups[0] === 'Children and Youth (0-17 years)')) {
+  /* If the 0-17 filter is not selected, don't show pediatric age group.*/
+  if(!selectedAgeGroups.includes('Children and Youth (0-17 years)')) {
     excludedFields.push(FilterableField.pediatricAgeGroup);
   }
+  
+  const selectedArboVirus = selectedFilters['pathogen'] ?? [];
+
+  if(selectedArboVirus.indexOf("DENV") == -1) {
```
HimanshuRanka, line 237:
```suggestion
  if(!selectedArboVirus.includes("DENV")) {
```

contexts/arbo-context/filter-update-steps/update-serotype-filter.tsx
```diff
@@ -0,0 +1,34 @@
...
+  export const updateSerotypeFilter = (
+    input: HandleArboFilterUpdateInput
+  ): HandleArboFilterUpdateOutput => {
+    if(input.action.payload.filter !== 'pathogen') {
+      return input;
+    }
+  
+    const selectedArboVirus = input.state.selectedFilters['pathogen'] ?? [];
+  
+    if(selectedArboVirus.indexOf("DENV") != -1) {
```
HimanshuRanka, line 18:
```suggestion
    if(!selectedArboVirus.includes("DENV")) {
```

HimanshuRanka (approved):
Other than the 2 small nitpicky changes this LGTM!

contexts/arbo-context/filter-update-steps/apply-new-selected-filters.tsx
```diff
@@ -14,10 +14,12 @@ export function filterData(
...
-      if (key === "pathogen"){
-        if(filters[key].length == 0){
-          return false;
-        }
+      if (key === "pathogen" && filters[key].length == 0 ){
+        return false;
+      }
+      /* Ignore pediatric age group for non pediatric age groups. */
+      if (key === "pediatricAgeGroup" && item["ageGroup"] !== "Children and Youth (0-17 years)"){
+        return true;
```
SeanKennyNF, line 22: This behaviour is inconsistent with the serotype filter. Are you sure this is what we want?

app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -229,10 +230,16 @@ export function Filters(props: FiltersProps) {
...
 
   const selectedAgeGroups = selectedFilters['ageGroup'] ?? [];
-
-  if(!(selectedAgeGroups.length === 1 && selectedAgeGroups[0] === 'Children and Youth (0-17 years)')) {
+  /* If the 0-17 filter is not selected, don't show pediatric age group.*/
+  if(!selectedAgeGroups.includes('Children and Youth (0-17 years)')) {
     excludedFields.push(FilterableField.pediatricAgeGroup);
   }
+  
+  const selectedArboVirus = selectedFilters['pathogen'] ?? [];
```
SeanKennyNF, line 238:
```suggestion
  const selectedArbovirus = selectedFilters['pathogen'] ?? [];
```
Couple other ones too with inconsistent capitalization.

SeanKennyNF (approved):
I think my comment on pediatric age group is definitely worth a look and feels like a bug but we can hit it in your next one if that's most convenient for you

contexts/arbo-context/filter-update-steps/apply-new-selected-filters.tsx
EliasPinno: I agree, it is inconsistent. But when we asked about how they wanted the serotype filter to behave, they said they wanted it to behave differently.

PR #238, opened 2024-03-24
Reduced minimum cluster size for map clusters from 10 to 2

### LINKED ISSUE #168: Reduce the minimum cluster size
Currently the jitter of pins is too low to see how many pins are there when a cluster size is small. 

we need to increase this slightly to make the pins more discerning. precise location is not super important as long as it is within a certain radius of the original place. 

We could wait to get some feedback from the users but I strongly believe this would be beneficial

SeanKennyNF:
Resolves #168. Reduced minimum cluster size for map clusters from 10 to 2. Looks fine imo.

image
Peek 2024-03-23 20-38


EliasPinno (approved):
I like it!

PR #242, opened 2024-03-24
Moved pathogen filter to the top of the filter section and off the map.

EliasPinno:
I've taken the legend off of the map and put in at the top of the scrollable filter bar.


components/customs/header.tsx
```diff
@@ -142,7 +142,7 @@ export const Header = () => {
           </h2>
         </Link>
       </div>
-      <NavigationMenu>
+      <NavigationMenu className="z-50">
```
EliasPinno, line 145: To make sure that he navigation top bar does not go "under" the top filter.

app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -215,10 +223,12 @@ export function Filters(props: FiltersProps) {
       }, // Include an empty object as payload
     });
   };
-
   if (filterData) {
     return (
       <div className={props.className}>
+        <div className="sticky p-2 z-10 top-0 w-full bg-white">
```
EliasPinno, line 229: bg-white to not have a gross looking half overlap with the elements below the fixed filter bar.

app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -195,7 +203,7 @@ export function Filters(props: FiltersProps) {
     {field: FilterableField.whoRegion, label: "WHO Region", valueToLabelMap: {}},
     {field: FilterableField.country, label: "Country", valueToLabelMap: {}},
     {field: FilterableField.antibody, label: "Antibody", valueToLabelMap: {}},
-    {field: FilterableField.pathogen, label: "Arbovirus", valueToLabelMap: {}},
+    {field: FilterableField.pathogen, label: "Arbovirus", valueToLabelMap: {}, className: "hidden"},
```
EliasPinno, line 206: I set this to hidden instead of deleting it. It is because this filter is used in the serotype PR in one of those apply filter phases. So we still need it in this list so it can be accessed later (I think).

SeanKennyNF (commented):
Sounds like we're going to make the arbovirus filter not sticky and potentially fix that bug you were talking about in standup. Just re-request whenever that's done and I'd be happy to take a peek :)

app/pathogen/arbovirus/dashboard/filters.tsx
HimanshuRanka: You do not need to make this sticky, you can just take this out of the scrollable filters part, and then have that be on top and then make the internal div with the other filters overflow-scroll.

app/pathogen/arbovirus/dashboard/filters.tsx
HimanshuRanka: Hmmm not sure if we want to do that. I am sure there is some way around this

HimanshuRanka (changes requested):
There is also a funky thing happening with the tooltip where the top and bottom have this transparent part. Not sure if that is realted to the changes here, but curious to see if it persists once u merge in changes from main

EliasPinno:
Too much hassle to fix these merge conflicts, and this changeset is kind of cursed anyways. Moved to another PR.

PR #248, opened 2024-03-29
ESM maps

HimanshuRanka:
I just realized I did this whole thing in the wrong branch LMFAO but all is well I think so

Anyway, in this PR I added in a new source + layer component that is specific to the ESM's -- I am aware this takes away reusability of the Pathogen Map component somewhat, but for the sake of speed that is ok. 

I also added in a filter that lets you select the esm. This is a single select filter and so I added in support for that as well. 

Now when u select this it filters the data accordingly and then overlays the esm on to the map.


app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -285,7 +296,7 @@ export function Filters(props: FiltersProps) {
           headerTooltipText="Filter on demographic variables, including population group, sex, and age group."
           allFieldInformation={demographicFilters}
           state={state}
-          filters={filterData.arbovirusFilterOptions}
+          filters={{...filterData.arbovirusFilterOptions, esm: ["zika", "dengue2015", "dengue2050"]}}
```
HimanshuRanka, line 299: since esms are not coming from the db, I needed to append them here

components/ui/pathogen-map/esm-maps.tsx
```diff
@@ -0,0 +1,46 @@
...
+const esmIds: Record<Esms, string> = {
+  zika: "zika_esm_source",
+  dengue2015: "dengue_2015_esm_source",
+  dengue2050: "dengue_2050_esm_source",
+};
+
+const esmUrls: Record<Esms, string> = {
+  zika: "mapbox://serotracker.7qe2zkgd",
+  dengue2015: "mapbox://serotracker.5wobnon2",
+  dengue2050: "mapbox://serotracker.9lzerqbc",
```
SeanKennyNF, line 16: Where did these urls come from?

components/ui/pathogen-map/esm-maps.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const selectedEsms = state.selectedFilters.esm[0];
+
+  return (
+    <Source
+      id={"esm_source"}
+      type="raster"
+      url={esmUrls[selectedEsms as Esms]}
+    >
+      <Layer
+        id={"zika_esm_source"}
```
SeanKennyNF, line 33: Did you mean to use `esmIds` here? I think `esmIds` is usused otherwise, right? Would be down to just give this some arbitrary id and remove `esmIds`.

components/ui/pathogen-map/pathogen-map.tsx
```diff
@@ -19,6 +19,8 @@ import {
 import { PathogenCountryHighlightLayer } from "./pathogen-country-highlight-layer";
 import { useCountryHighlightLayer } from "./use-country-highlight-layer";
 import isEqual from "lodash/isEqual";
+import { EsmMapSourceAndLayer } from "./esm-maps";
+import { layer } from "@fortawesome/fontawesome-svg-core";
```
SeanKennyNF, line 23: Unused import?

components/ui/pathogen-map/pathogen-map.tsx
```diff
@@ -1,7 +1,7 @@
 import { MapResources } from "@/app/pathogen/arbovirus/dashboard/(map)/map-config";
 import { getEsriVectorSourceStyle } from "@/utils/mapping-util";
 import { useState, useEffect, MutableRefObject, useRef, useMemo } from "react";
-import { Map, NavigationControl } from "react-map-gl";
+import { Layer, Map, NavigationControl, Source } from "react-map-gl";
```
SeanKennyNF, line 4: Unused import?

SeanKennyNF (approved):
Super cool man. Just to clarify, black means suitable environment, right? I wonder if there's something lighter and less opaque we could maybe use so that the country highlight layer is still visible underneath it :thinking:

components/ui/pathogen-map/esm-maps.tsx
HimanshuRanka: I got the rasters from yannik, converted them to 8bits and then uploaded them to mapbox giving us these links

HimanshuRanka:
> use

Yeah I wanted to do something like that too but Harriet and Mairead said it looks good as is for now and wanted to keep it as such. So did not mess with it

SeanKennyNF (approved):

PR #253, opened 2024-03-30
Made the dashboards responsive

HimanshuRanka:
There are still many issues that will just take a lot more planning and involevment to solve. Especially with the visualizations and data table. Idk if I am just being silly and there is an easy way to solve it but either way have thrown a little disclaimer on there. 

It still kind of usable but we will need to make design changes across the website and it will take more time and desiging. Let me know if there is something that looks too bad and I have to fix it (other than the viz cause they are actually asses and do not comply)

### DIFF
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -28,7 +28,7 @@ export default function RootLayout({
         <link rel="alternate icon" href="/SerotrackerLogo.png" />{" "}
         {/* Fallback for older browsers */}
       </head>
-      <body className={clsx(inter.className, "text-black no-scrollbar")}>
+      <body className={clsx(inter.className, "text-black no-scrollbar overflow-hidden")}>
         <ThemeProvider>
           <Header />
           <main className={"h-full-screen w-screen bg-foreground"}>
--- a/app/pathogen/arbovirus/dashboard/page.tsx
+++ b/app/pathogen/arbovirus/dashboard/page.tsx
@@ -14,7 +14,7 @@ export default function ArbovirusDashboardPage() {
 
   return (
     <div className="col-span-12 grid gap-0 grid-cols-12 grid-rows-4 lg:grid-rows-2 row-span-2 grid-flow-col w-screen overflow-hidden border-box">
-      <div className="overflow-y-scroll col-span-12 row-span-1 lg:col-span-2 lg:row-span-4 h-full border-y border-b border-background lg:border-b-0">
+      <div className="overflow-y-scroll col-span-12 row-span-1 lg:col-span-2 lg:row-span-4 h-full border-b border-background lg:border-b-0">
         <Filters
           className="p-4 border-background"
         />
@@ -22,7 +22,6 @@ export default function ArbovirusDashboardPage() {
       <div
         className={"overflow-y-scroll snap-y scroll-smooth row-span-3 col-span-12 lg:col-span-10 lg:row-span-2 px-4"}
       >
-        <ArboBanner/>
         {!isLargeScreen && (
           <section
           className="w-full h-fit relative row-span-2 rounded-md mt-4 border border-background p-4"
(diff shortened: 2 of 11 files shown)


app/pathogen/arbovirus/dashboard/page.tsx
```diff
@@ -5,18 +5,29 @@ import { VisualizationsSection } from "./(visualizations)/visualizations-section
...
       <div
-        className={"overflow-y-scroll snap-y scroll-smooth col-span-10 row-span-2 px-4"}
+        className={"overflow-y-scroll snap-y scroll-smooth row-span-3 col-span-12 lg:col-span-10 lg:row-span-2 px-4"}
       >
+        {!isLargeScreen && (
+          <section
+          className="w-full h-fit relative row-span-2 rounded-md mt-4 border border-background p-4"
+        >
+          We suggest viewing our dashboard on a larger screen or in landscape mode for the best experience. Support for smaller devices is coming soon!
+        </section>
```
SeanKennyNF, line 29: Such a good idea :heart:

hooks/useIsLargeScreen.ts
```diff
@@ -0,0 +1,19 @@
...
+    };
+
+    checkScreenSize(); // Check on initial render
+
+    window.addEventListener('resize', checkScreenSize);
+    return () => window.removeEventListener('resize', checkScreenSize);
+  }, []);
+
+  return isLargeScreen;
+};
```
SeanKennyNF, line 19:
```suggestion
import { useMemo } from 'react';

// Checks if the screen width is greater than 1024px
export const useIsLargeScreen = () => {
  const isLargeScreen = useMemo(() => window.innerWidth > 1024, [window.innerWidth])

  return isLargeScreen;
};
```
I believe this is the same performance-wise and doesn't add a screen listener.

app/pathogen/arbovirus/dashboard/page.tsx
SeanKennyNF, line 29: I actually wonder if we could remove the whole `isLargeScreen` thing and just conditionally put `display: hidden` on this with a tailwind breakpoint for screen sizes under `lg` (The tailwind class is called `hidden` I think?). That way we're doing it all in css.

I feel like having stuff with `display: hidden` makes it not occupy space in the grid but I could be wrong?
Just an idea if you haven't given that a go already hahaha

SeanKennyNF (dismissed):
Yeah I agree it's still pretty rough for mobile users but a massive improvement anyways. Left a couple suggestions :)

app/pathogen/arbovirus/dashboard/page.tsx
SeanKennyNF: If your idea with the table was to make the column not sticky on certain screen sizes, I feel like that could probably be done in the css of the fixed column so I feel like that's NABD either.

EliasPinno (dismissed):
Seems good. Yeah this will be a fairly involved process to fully fix.

SeanKennyNF (commented):
Failing build feel free to re-request though :)

app/layout.tsx
```diff
@@ -28,7 +28,7 @@ export default function RootLayout({
         <link rel="alternate icon" href="/SerotrackerLogo.png" />{" "}
         {/* Fallback for older browsers */}
       </head>
-      <body className={clsx(inter.className, "text-black no-scrollbar")}>
+      <body className={clsx(inter.className, "text-black no-scrollbar overflow-hidden")}>
```
SeanKennyNF, line 31: Commenting on random line.

This is a regression.

image

SeanKennyNF, line 31: Commenting on a random line, you can also download csvs on the table with WHO regions and age groups. I'm assuming that one's fine?

SeanKennyNF (approved):
Cool, big improvement imo.

- I feel like we should update the FAQ if we're removing the download CSV option.
- I would consider the comment I left about simplifying the hook. It seems more resilient than a window listener.

app/layout.tsx
HimanshuRanka: regression?

HimanshuRanka:
> Cool, big improvement imo.
> 
> * I feel like we should update the FAQ if we're removing the download CSV option.
Bet I will add this to one of the other tickets


> * I would consider the comment I left about simplifying the hook. It seems more resilient than a window listener.
I tired to do that but next js tried to render the pages on the server and I was not able to figure out why, and it crashed (build issue) so I left it for now and made a little note so we can come back to it (works, but is not the most performant i guess)

Issue #247, opened 2024-03-28
Make the Ages table full screen width, update download button to download data


HimanshuRanka:
We want to make the table full screen (so takes up both rows of the grid) at the very bottom of the page.

We want to make it so that if someone were to download the data it comes out as a single csv file with pathogen as a column
pathogen, whoregion, age 1, age 2
Zika, who1, data age 1, data age 2
Zika, who2, data age 1, data age 2
DENV, who1, data age 1, data age 2 
DENV who2, data age 1, data age 2

And finally we want to add in a way for people to know how many pages there are
currently it is not clear what the next and prev buttons are doing.

SeanKennyNF:
Acknowledging that I see this. Will get on it :)

PR #254, opened 2024-03-31
Added a cleaner arbovirus selection method on "Median seroprevalence by WHO region and age group"

SeanKennyNF:
Part of resolving #247. Basically made it a dropdown instead of a pages thing. Looks and works way better imo.

Peek 2024-03-30 19-03


app/pathogen/arbovirus/dashboard/(visualizations)/median-seroprevalence-by-who-region-and-age-group-table.tsx
```diff
@@ -169,9 +170,22 @@ export const MedianSeroprevalenceByWhoRegionAndAgeGroupTable = () => {
   return (
     <div className="p-2">
       <div className="flex justify-between mb-2 ignore-for-visualization-download">
-        <h2 className="text-center">
-          {convertArboSFtoArbo(currentlySelectedArbovirus)}
-        </h2>
+        <DropdownMenu>
+          <DropdownMenuTrigger asChild>
+            <Button variant="outline" size="sm" className="mx-2 whitespace-nowrap text-white ignore-for-visualization-download">
```
SeanKennyNF, line 175: Yeah this looks pretty bad in the downloaded file so I excluded it. The downloaded file used to say which arbovirus it was for but now it doesn't anymore. I'm going to sleep on this, I think there's probably a better solution hiding somewhere.

EliasPinno (approved):
Didn't break anything: good work!

PR #255, opened 2024-03-31
Made the median seroprev by who region and age group table occupy the full width of the page

SeanKennyNF:
Part of resolving #247.

image


EliasPinno (approved):
My IQ was small, but this still makes sense.

PR #256, opened 2024-03-31
Re-opening of moving the map arbovirus filter to the filter component on the left side of the map.

### LINKED ISSUE #186: Remove the "Arbovirus" filter from the filters on the left on the redesigned dashboard page when the map is in view.
Created from https://github.com/serotracker/dashboard/pull/183#discussion_r1493831304

> So an interesting consequence of what I've done is that the pathogen filter will always be shown since I'm no longer tracking which section is in view. If removing the pathogen filter when you're looking at the map is a feature we're interested in having on the new dashboard, let's make a separate ticket for that because it's a bit more involved now.

When the map is visible, the "Arbovirus" filter on the left of the map (circled in red on the left) should disappear because it is redundant with the arbovirus checkboxes present on the right side of the map (circled in red on the right). The Arbovirus filters should be visible again once someone scrolls to a different section of the dashboard such as the table or the visualizations so it's not sufficient to just look at the url param and we need something that determines whether or not the map is in view I'd imagine.

Screenshot from 2024-02-19 14-56-00

### LINKED ISSUE #224: Change order of filters
1. Sampling Start Date, Sampling End Date (most important)
2. Study Information
    - WHO Region
    - UN Region
    - Country
3. Demographic    
    - Age Group
    - Sex
    - Sample Frame
4. Pathogens
5. Test Information (new heading)    
    - Assay
    - Assay Producer
    - Antibody
6. Reset Filters

### LINKED ISSUE #228: Incorporate what shading means into map legend
Similar to SeroTracker, incorporate into the Arboviruses legend

Image

### LINKED ISSUE #234: Add a banner in arbo dashboard just above the map
Image

This banner should also contain the blue button to download the csv which currently resides near the data tables

EliasPinno:
Resolving merge conflicts was very annoying for what was actually a very small change set. New cleaner version of the changes, which also includes some other ticket resolutions. 

Overview of everything this PR does:

Re orders the filters as requested by #224.
Takes the arbovirus filter off the map and puts it on the filters side bar as requested by #186 
Fixes a bug where that filter was not in sync with the state of the filters on the side of the bar.
Adds a banner which downloads a CSV at the very top of the page as requested by #234 
Adds a legend where the arbovirus filter used to be on the map as requested by #228 

Closes #228 
Closes #186
Closes #234
Closes #224


app/pathogen/arbovirus/dashboard/(banner)/ArboBanner.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const ArboBanner = ({className}: ArboBannerProps) => {
+    const state = useContext(ArboContext);
+    
+    const downloadData = () => {
+        const csvConfig = mkConfig({
+            useKeysAsHeaders: true,
+            filename: "arbotracker_filtered_dataset",
+          });
+          let csv = generateCsv(csvConfig)(state.filteredData);
+          download(csvConfig)(csv);
```
EliasPinno, line 21: !!! BEHAVIOR CHANGE !!!

I want to note something important here. The old csv download button lived on the table component, and only saved the columns you were looking at. This one is not on the table. So it can't see the columns you have selected -> nor should it. It would be very confusing if a download button not on the table was effected by the table. It still does only download the portion of the data you have filtered though (which is also confusing? maybe the button should say "download currently selected data?").

There are a few things we can do about this. Right now I left the original download button is untouched until we decide what to do: we could relabel it "download table csv" or something like that and have both? Or maybe we don't care about csv column filtering and we can just delete the other tables csv download button and not care about the missing columns.

EliasPinno: Things I'd love from a reviewer: what should the button(s) say? And should there be button(s) or just the one on the banner?

app/pathogen/arbovirus/dashboard/(banner)/ArboBanner.tsx
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    return (
+      <div className={className}>
```
EliasPinno, line 25: I followed the example of the other components and put a lot of the styling choices for this component on the outside in page.tsx, but maybe this is confusing? For example the side by side text button layout here comes from the outside of this component? Maybe a bad idea.

app/pathogen/arbovirus/dashboard/(map)/MapArbovirusFilter.tsx
```diff
@@ -1,18 +1,17 @@
...
 import { pathogenColorsTailwind } from "./ArbovirusMap";
 import { Button } from "@/components/ui/button";
+import SectionHeader from "@/components/customs/SectionHeader";
 
 interface MapArbovirusFilterProps {
-  records: unknown[];
   className?: string;
+  state: ArboContextType;
+  data: any;
 }
```
EliasPinno, line 11: This component now takes in just the exact same data sources as the other filters: solves the sync problem WITHOUT me having to do any really annoying stuff.  Always good.

app/pathogen/arbovirus/dashboard/(map)/MapArbovirusFilter.tsx
```diff
@@ -100,8 +100,8 @@ export const MapArbovirusFilter = ({ records, className }: MapArbovirusFilterPro
             </div>
           );
         })}
-        <Button className="mt-2" variant={'ghost'} onClick={clearAllHandler}>Clear all</Button>
-      </CardContent>
-    </Card>
+        <Button className="mt-2" variant={'ghost'} onClick={clearAllHandler}>Clear all viruses</Button>
+      </div>
+    </div>
```
EliasPinno, line 105: I have left this button in for now, but if requested I can delete it. I think it kind of makes sense to leave it in?

app/pathogen/arbovirus/dashboard/(map)/MapShadingLegend.tsx
```diff
@@ -0,0 +1,36 @@
...
+interface MapShadingLegendProps {
+    className?: string;
+  }
+
+
+export const MapShadingLegend = ({className}: MapShadingLegendProps) => {
+    const shadingLegendElements: {color: string, description: string}[] = [
+        {color: MapSymbology.CountryFeature.HasData.Color, description: "Seroprevalence estimates"},
+        {color: MapSymbology.CountryFeature.Default.Color, description: "No Seroprevalence estimates"}
+    ]
```
EliasPinno, line 14: I didn't see any "Non Applicable" color on map anywhere. So to me it only made sense for this legend to have two entries.

app/pathogen/arbovirus/dashboard/(table)/columns.tsx
EliasPinno, line 1: Because I squash merged this shows up as diff I wrote? This is confusing, as these changes are already merged into main: I don't know why this file is showing up in the diff.

app/pathogen/arbovirus/dashboard/filters.tsx
```diff
@@ -288,10 +319,15 @@ export function Filters(props: FiltersProps) {
           filters={filterData.arbovirusFilterOptions}
           data={data}
         />
+        <MapArbovirusFilter 
+          data={data} 
+          state={state} 
+          className={"p-0"} 
+        />
```
EliasPinno, line 326: This is what the ticket requests, but this filter looks awful here. Much better at the top of the filters bar in my opinion. Give me permission and I'll move it.

EliasPinno: I moved it!

app/pathogen/arbovirus/dashboard/(banner)/ArboBanner.tsx
HimanshuRanka: These are really really good points actually. I think for now I am happy to leave it as is. I will elevate the concern to the team and we can further test it with users to see what they expect. 

One super quick fix is to add a tooltip on hover of hte button explaining it will download filtered data. 

I am ok ignoring the columns tbh, it is an easy fix for whoever downloads the data. 

Another possible solution is to simply create a popup on click. This has a couple radio buttons or something like this that lets them select whether they want all data or filtered data, alternatively it could be a dropdown... much to consider.

Happy leaving it as is for now as it is not a "breaking" change and the user will still have access to all the data if the need arises.

app/pathogen/arbovirus/dashboard/(banner)/ArboBanner.tsx
HimanshuRanka: Yeah I agree this is not something we want to do --> would've made sense if it was a customizable or reusable component both of which it is not. Can create a ticket to quickly address that post launch -- ok for now and good callout!

app/pathogen/arbovirus/dashboard/(banner)/ArboBanner.tsx
```diff
@@ -0,0 +1,37 @@
+"use client";
```
HimanshuRanka, line 1: another nit: We do not need a directory called (banner) since no other files will go in here. You can lift the file up to the dashboard directory and it is essentially the same

app/pathogen/arbovirus/dashboard/(map)/MapArbovirusFilter.tsx
HimanshuRanka: yeah lets leave it in

app/pathogen/arbovirus/dashboard/(map)/MapShadingLegend.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+    return (
+        <Card className={className}>
```
HimanshuRanka, line 17: Similar to above, should localize styling -- Let us preserve using the className prop ONLY for reusable components

HimanshuRanka (approved):
LGTM! comments are about stuff that can be addressed later!  Please create tickets for outstanding tasks before merging tho <3

HimanshuRanka:
image

WAIT why is the ESM filter not showing up?

app/pathogen/arbovirus/dashboard/(banner)/ArboBanner.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+    return (
+      <div className={className}>
+        <p className="text-sm">
+            ArboTracker is a dashboard and platform for Arbovirus serosurveys. 
+            We conduct an ongoing systematic review to track serosurveys 
+            (antibody testing-based surveillance efforts) around the world and 
+            visualize findings on this dashboard.
+        </p>
+        <Button className="w-[30%]" onClick={downloadData}>
```
HimanshuRanka, line 32:
```suggestion
        <Button className="w-[30%] bg-background" onClick={downloadData}>
```

HimanshuRanka:
Whoops can u also remove the download csv button from the table?

HimanshuRanka (changes requested):
Ok a couple things do need to be changed before we can merge

app/pathogen/arbovirus/dashboard/(banner)/ArboBanner.tsx
EliasPinno: Done!

app/pathogen/arbovirus/dashboard/(map)/MapShadingLegend.tsx
EliasPinno: I localized it!

app/pathogen/arbovirus/dashboard/ArboBanner.tsx
```diff
@@ -22,14 +19,14 @@ export const ArboBanner = ({className}: ArboBannerProps) => {
...
-      <div className={className}>
+      <div className="w-full border border-background rounded-md items-center p-2 mt-4 flex justify-between">
         <p className="text-sm">
             ArboTracker is a dashboard and platform for Arbovirus serosurveys. 
             We conduct an ongoing systematic review to track serosurveys 
             (antibody testing-based surveillance efforts) around the world and 
             visualize findings on this dashboard.
         </p>
-        <Button className="w-[30%]" onClick={downloadData}>
+        <Button className="w-[30%] bg-background" onClick={downloadData}>
```
HimanshuRanka, line 29:
```suggestion
        <Button className="w-[30%] bg-background hover:bg-backgroundHover" onClick={downloadData}>
```

HimanshuRanka (approved):

PR #257, opened 2024-03-31
The estimate id column is only fixed in the table at the lg breakpoint and above (responsive design for mobile)

SeanKennyNF:
Part of #226.

Peek 2024-03-31 15-37


EliasPinno (approved):
Makes sense. Looks good.

PR #259, opened 2024-03-31
Resolved #215

### LINKED ISSUE #215: Using the table option in the header when on arbo dashboard does not work
When we use the header to naviagate to the table in arbo while already on the arbo page, the header collapses and the map begins to autoscroll but the mouse ends up landing on top of a map modal, which triggers on hover event which pauses the scroll, and when we move the mouse away the page springs back to the top. 

This is an issue and should not be happening

Additionally we should add in a triangle of activation to our header dropdown so it works more smoothly

SeanKennyNF:
Resolves #215

Peek 2024-03-31 16-32


EliasPinno (approved):
LGTM.

Issue #271, opened 2024-04-06
Feedback 2


HimanshuRanka:
AMAZING WORK! I appologize if this has already been noted elsewhere:

- The word "trend" is spelled incorrectly, "Your Go to Source for COVID-19 and Arbovirus Seroprevalence Data
We synthesize findings from thousands of COVID-19 and Arbovirus seroprevalence studies worldwide, collect and standardize the data we extract from them, provide useful analytics on the data and extract monthly insights relating to tredns and patterns we find in the data"

- I know this is not critical, but for those that do not know what ESM is, maybe a small legend to show what the shading represents?

-Not sure if this is my screen, but there is a lot of green space at the bottom
Zoom image.png
image.png 90.3 KB [View full-size](https://storage.3.basecamp.com/5061858/blobs/04359930-f355-11ee-b869-b2d1a80aa527/download/image.png) [Download](https://storage.3.basecamp.com/5061858/blobs/04359930-f355-11ee-b869-b2d1a80aa527/download/image.png?attachment=true)

- the dates look very squished:
Zoom image.png
image.png 156 KB [View full-size](https://storage.3.basecamp.com/5061858/blobs/1b48190e-f355-11ee-a3d7-1e0199a71503/download/image.png) [Download](https://storage.3.basecamp.com/5061858/blobs/1b48190e-f355-11ee-a3d7-1e0199a71503/download/image.png?attachment=true)

PR #275, opened 2024-04-07
Made some changes based on feedback from the team

HimanshuRanka:
Addresses #270 and #271

### DIFF
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -28,7 +28,7 @@ export default function RootLayout({
         <link rel="alternate icon" href="/SerotrackerLogo.png" />{" "}
         {/* Fallback for older browsers */}
       </head>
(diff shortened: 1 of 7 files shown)


app/layout.tsx
```diff
@@ -28,7 +28,7 @@ export default function RootLayout({
         <link rel="alternate icon" href="/SerotrackerLogo.png" />{" "}
         {/* Fallback for older browsers */}
       </head>
-      <body className={clsx(inter.className, "text-black no-scrollbar overflow-hidden")}>
```
HimanshuRanka, line 31: Initially added this in to avoid that large chunk of greenspace at the bottom, but then with out #id based navigation it broken the header and scroll.

So gonna leave for now as it is ugly but functionaly and create a ticket to resolve

SeanKennyNF (approved):

PR #279, opened 2024-04-07
Made WHO region and age group table more dynamic.

### LINKED ISSUE #262: Make the table viz for median age stuff responsive
Only parts of the table with data should display instead of having a bunch of tables and a bunch of NA values. 

For example if I select India then there should only be one row and 3 table options (since we do not have all the arboviruses there)

This will make the UX much cleaner

SeanKennyNF:
Made WHO region and age group table more dynamic. Resolves #262 

Peek 2024-04-07 13-06

Peek 2024-04-07 13-08

Reflected in the csv which has less columns:
[median-seroprevalence-by-who-region-and-age-group(2).csv](https://github.com/serotracker/dashboard/files/14898590/median-seroprevalence-by-who-region-and-age-group.2.csv)


SeanKennyNF:
Oh cool, I actually shipped this as part of my other PR. Closing and resolving issue.

Issue #268, opened 2024-04-03
Edits to About page


harrietware:
Expect further edits to the About page copy as our collaborating partners review!

See edits here:
https://docs.google.com/document/d/1PZTfERqHfF_HQk4w3nBFIDHnPJgDoBBy6FpEOG6jSMw/edit

harrietware:
In the ArboTracker FAQ, "How often is ArboTracker data updated?" is repeated twice. Can we change the first instance to "Where does ArboTracker data come from?"

PR #284, opened 2024-04-14
Final touches to the "SeroTracker Team" page for now. A bit of style changes to the headings.

SeanKennyNF:
Last bit of changes for #268.

image

### DIFF
--- a/app/about/the-team/page.tsx
+++ b/app/about/the-team/page.tsx
@@ -64,6 +64,22 @@ const TeamMemberInfoCard = (props: TeamMemberInfoCardProps) => {
   )
 }
 
+interface TeamCardHeaderProps {
+  label: string;
+}
+
+const TeamCardHeader = (props: TeamCardHeaderProps) => {
+  if(props.label === 'ArboTracker Collaborating Partners') {
+    return <h2 className="my-4 pb-2 border-b border-background"> {props.label} </h2>;
+  }
+
+  if(props.label === 'SeroTracker Alumni') {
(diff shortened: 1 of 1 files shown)


app/about/the-team/page.tsx
```diff
@@ -64,14 +64,30 @@ const TeamMemberInfoCard = (props: TeamMemberInfoCardProps) => {
...
+  if(props.label === 'ArboTracker Collaborating Partners') {
+    return <h2 className="my-4 pb-2 border-b border-background"> {props.label} </h2>;
+  }
+
+  if(props.label === 'SeroTracker Alumni') {
+    return <h3 className="my-4 border-b mt-32 border-background"> {props.label} </h3>;
+  }
+
+  return <h3 className="my-4 border-b border-background"> {props.label} </h3>;
+}
```
SeanKennyNF, line 81: Yeah I know this is gross and I'm tying the names in Airtable to our frontend. I think this is fine because I don't know if we're going to want to do this kind of thing often enough and keep this change permanently enough for me to do something more fancy and time consuming like putting a new field in Airtable.

This is fine imo, a bit silly but pretty low consequences on a piece of code that might die at any time.

HimanshuRanka: Agreed! could we add that in a comment in the code directly so if we ever come back to it we know hwy it does not need to be improved

EliasPinno (approved):

app/about/the-team/page.tsx
SeanKennyNF: For sure, let me make a comment on my next PR, I've got a couple things I want to hit for Elias in my next one too :)

PR #285, opened 2024-04-14
Improved the usability of the 95% CI graph.

### LINKED ISSUE #261: Make 95% CI graph more informative
For this viz, move the helper text to the center, make it non responsive (Mentions its only works for one pathogen and 2 countries --> very frustrating to have to do both seperately without knowing) 

Additionally allow them to filter for the values right there in the visualization -- makes for a much smoother ux... maybe.

SeanKennyNF:
Resolves #261. Improves the usability of the 95% CI graph.

Peek 2024-04-13 20-42


app/pathogen/arbovirus/dashboard/(visualizations)/chart-arbovirus-dropdown.tsx
```diff
@@ -0,0 +1,95 @@
+import { useContext, useMemo, useState, useCallback } from "react";
```
SeanKennyNF, line 1: This is mostly copy-paste from the WHO region and Age group table. I wanted to make that code reusable across both components and quarantine the gross things I had to do to make this work with all of the edge cases.

EliasPinno (approved):

PR #286, opened 2024-04-14
Removed the green bar at the bottom of the screen on Chrome

### LINKED ISSUE #276: Figure out what is up with the greenspace at the bottom of the arbo page
I know this is related to the filters component and something to do with how we are handling overflow (not correctly) but have not been able to figure out why yet

SeanKennyNF:
Resolves #276.

Firefox:
Peek 2024-04-13 20-55

Chrome:
Peek 2024-04-13 20-56

### DIFF
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -28,7 +28,7 @@ export default function RootLayout({
         <link rel="alternate icon" href="/SerotrackerLogo.png" />{" "}
         {/* Fallback for older browsers */}
       </head>
-      <body className={clsx(inter.className, "text-black no-scrollbar")}>
+      <body className={clsx(inter.className, "text-black no-scrollbar overflow-y-hidden")}>
         <ThemeProvider>
           <Header />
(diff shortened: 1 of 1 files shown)


HimanshuRanka:
I tried this as well, but it breaks header navigation...

HimanshuRanka (changes requested):
If we use header navigation it blocks the header and we can no longer scroll up to it, we need something more involeved.

This is definitely related to filters (removing the filter component fixes the issue) not sure why though

SeanKennyNF:
> If we use header navigation it blocks the header and we can no longer scroll up to it, we need something more involeved.
>
> This is definitely related to filters (removing the filter component fixes the issue) not sure why though

Yeah I'm actually trying to hit both this one and the header one together. Yeah you're right these are kind of related.

SeanKennyNF:
Actually yeah let's close this guy. I'm opening something a bit more comprehensive :+1:

PR #289, opened 2024-04-27
Added a button to view a tutorial for the dashboard.

### LINKED ISSUE #287: Add dashboard tutorial video to dashboard
Link somewhere near top of page?

https://drive.google.com/file/d/1cLslLkwI57f2oKI45utkKdOtlW9ZCWpm/view?usp=drive_link

SeanKennyNF:
Resolves #287.

Peek 2024-04-27 11-52

### DIFF
--- a/app/pathogen/arbovirus/dashboard/ArboBanner.tsx
+++ b/app/pathogen/arbovirus/dashboard/ArboBanner.tsx
@@ -42,7 +42,7 @@ export const ArboBanner = () => {
         }}>
           Get Citation for CSV
         </Button>
-        <Link href="https://drive.google.com/file/d/1cLslLkwI57f2oKI45utkKdOtlW9ZCWpm/view" target="__blank" rel="noopener noreferrer">
+        <Link href="https://drive.google.com/file/d/1cLslLkwI57f2oKI45utkKdOtlW9ZCWpm/view?usp=sharing" target="__blank" rel="noopener noreferrer">
           <Button className="w-[30%] bg-background hover:bg-backgroundHover ml-2">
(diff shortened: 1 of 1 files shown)


HimanshuRanka (approved):

HimanshuRanka (changes requested):
I have some design feedback for this

HimanshuRanka:
Just wanted to throw a bit of a frontend design perspective curveball at you, but instead of using a button on the arbo dash that links to the gdrive, can you design an integration of the video into the landing page that looks good, and then implement it? Maybe use figma, or like handdrawn with imges or whatever is easiest to relay the design idea.

We want the video to be playable straight on the website. Should be seen right on the landing page (maybe this would change not sure but having it right ther is good visibility so people can learn the platform before going)

Options: 
- put it in the hero section
- put it below the hero section
- put it further down when a user scrolls
- put it in a pop up (i do not like popups usually but consider it) 

Let me know what u decide on and why!

SeanKennyNF:
I've thrown together 2 little things for us to look at.

https://github.com/serotracker/dashboard/pull/297 and https://github.com/serotracker/dashboard/pull/298

Both are designs where the video is on the landing page. I like the first one a lot more than the second one but if we think the first one is a bit too busy I think that's a reasonable point and that's why the second one is there too.

I feel like a reasonable option ould be to have two columns, one with our existing stuff and another with a small blurb and the video but I figured these two options are probably a good start.

Are we feeling either of these options? Do you feel like one of them is more of a step in the right direction than the other?

SeanKennyNF:
Continued in https://github.com/serotracker/dashboard/pull/297

PR #297, opened 2024-04-28
Tutorial video in Landing Page

SeanKennyNF:
image


app/page.tsx
```diff
@@ -85,12 +87,12 @@ export default async function Home() {
             backgroundImage: `url(${SeroMap.src})`,
           }}
         >
-          <div className="flex flex-col w-full rounded-md p-4 mb-4 min-h-1/2 h-fit text-background bg-white/90">
-            <div className="w-full h-fit mb-2 lg:mb-0">
+          <div className="grid grid-cols-4 w-full rounded-md p-4 mb-4 min-h-1/2 h-fit text-background bg-white/90">
```
SeanKennyNF, line 90: This obviously looks bad on smaller screens. I'll fix the responsiveness if we decide this is a direction we're interested in going in.

HimanshuRanka:
This is the better alt. But can we move the buttons underneath the text and have the video just take up the right side of the screen instead

SeanKennyNF:
Is this what you had in mind @HimanshuRanka?

image

SeanKennyNF:
I think this looks worse than the one above but throwing it in as well.

image

SeanKennyNF:
We're interested in moving the buttons under the text and making the video smaller than the text like heightwise

harrietware:
The video tutorial is only for ArboTracker, not all of SeroTracker - could we make that clearer in the text or title if we keep the placement on the landing page?

SeanKennyNF:
> The video tutorial is only for ArboTracker, not all of SeroTracker - could we make that clearer in the text or title if we keep the placement on the landing page?

For sure Harriet! This is the copy I could come up with but if we want something better, just let me know!

"Check our dashboards using the buttons to the right or view our short tutorial on how to use our ArboTracker dashboard!"

Talked with Himanshu about this one, I just made this a bit little bit better on mobile :)

SeanKennyNF:
@harrietware + @HimanshuRanka 
Do you folks have any additional feedback here?
Are we comfortable moving this forwards?

HimanshuRanka (approved):

PR #298, opened 2024-04-28
Tutorial video in Landing Page Alt #2

SeanKennyNF:
image

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -85,6 +85,7 @@ export default async function Home() {
             backgroundImage: `url(${SeroMap.src})`,
           }}
         >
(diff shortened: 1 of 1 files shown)


app/page.tsx
```diff
@@ -148,6 +149,16 @@ export default async function Home() {
               )}
             </div>
           </div>
+          <div className="flex flex-col w-1/2 rounded-md p-4 mb-4 min-h-1/2 h-fit text-background bg-white/90">
```
SeanKennyNF, line 152: `w-1/2` bad for responsive design, I'll make this proper if we decide it's the direction we want to go in.

PR #308, opened 2024-05-06
Added published study count by GBD region visualization

SeanKennyNF:
Added published study count by GBD region visualization

image
image


app/pathogen/arbovirus/dashboard/page.tsx
```diff
@@ -14,7 +14,7 @@ export default function ArbovirusDashboardPage() {
       filtersComponent={ArbovirusFilters}
```
HimanshuRanka, line 14: +1 on this change, looks super clean

app/pathogen/arbovirus/visualizations/page.tsx
```diff
@@ -1,45 +1,16 @@
 "use client"
```
HimanshuRanka, line 1: Only note here is that it probably makes sense to merge the visualization and (visualization) directories. I understand why they would be seperate technically but they are never used without each other and hence lets centralize it - and avoid confusion going forward

HimanshuRanka (approved):
Looks great to me! This is amazing work sean thank you!

app/pathogen/arbovirus/visualizations/page.tsx
SeanKennyNF: Let me see what we can do once there's less open PRs hahaha :)

PR #309, opened 2024-05-11
Created "Cumulative Number Serosurveys Published Over Time" graph.

### LINKED ISSUE #221: Cumulative Number Serosurveys Published Over Time
Create viz as seen on serotracker.com

SeanKennyNF:
Resolves #221 

I'm aware of the problem with it being cut off a little bit at the bottom, let's get the graph down and then do little tweaks like fixing that after.

image

image


app/pathogen/arbovirus/dashboard/(visualizations)/recharts.tsx
```diff
@@ -644,124 +644,3 @@ export function Top10CountriesByPathogenStudyCount() {
     </div>
   );
 }
-
-//Cumulative Study count over time by smaple frame
```
SeanKennyNF, line 648: This is the arbo visualization but I changed it to use my new component.

EliasPinno (approved):

PR #331, opened 2024-05-28
Fixed the bugs with incorrect data appearing for the clusters.

### LINKED ISSUE #269: Fix map filter bugs
Noticed a couple of bugs when using the Arbovirus checkboxes to filter the map
- Screenshot 1: Only Zika and Dengue selected but the # of estimates clustered show other viruses 
- Screenshot 2: ESM Dengue 2015 selected and map looks good - then ELISA filter applied and # of estimates clustered show other viruses

In both these cases when you zoom in on the clusters, it displays the correct data

Image

Image

SeanKennyNF:
Resolves #269. The map flickers a little bit when the filters change even if the cluster hasn't changed it's data. I think the intent of the original caching code (which was causing this problem) was to remove the flicker and to save generating elements we don't have to. This saves generating elements we don't have to but still has problems with flickering. Production also has problems with flickering, I think we should do something more in-depth to remove that flickering but I'm just fixing the bug for now.

Peek 2024-05-27 18-24

### DIFF
--- a/app/pathogen/arbovirus/dashboard/(map)/arbo-map-cluster-utils.tsx
+++ b/app/pathogen/arbovirus/dashboard/(map)/arbo-map-cluster-utils.tsx
@@ -232,8 +232,6 @@ export function computeClusterMarkers(props: {
   const newMarkers: MarkerCollection<ArbovirusMarkerProperties> = {};
   // for every cluster on the screen, create an HTML marker for it (if we didn't yet),
   // and add it to the map if it's not there already
-  console.log('props.features', props.features)
-
   for (const feature of props.features) {
     // A geometryCollection does not have a coordinate property. Instead it has a collection of geometries.
     // I do not believe we will be dealing with this use case atm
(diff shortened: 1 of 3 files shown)


app/pathogen/arbovirus/dashboard/(map)/arbo-map-cluster-utils.tsx
```diff
@@ -217,22 +241,37 @@ export function computeClusterMarkers(props: {
...
+        else {
+          const marker = createClusterPointMarker({
+            element: element,
             coords: [coords[0], coords[1]],
             id: id,
             map: props.map,
           });
-        }
-        if (marker) {
-          newMarkers[id] = marker;
```
SeanKennyNF, line 235: The culprit

HimanshuRanka: wait tell me more

HimanshuRanka (approved):

PR #342, opened 2024-06-03
Reach filter parity with serotracker.com

### LINKED ISSUE #339: Reach filter parity with serotracker.com
https://3.basecamp.com/5061858/buckets/30589330/messages/7438758503

> Filters: Please include all filters that are on the original SARS-CoV-2 dashboard. Source Type is listed twice, once under Test Information.

- Fix Source type being listed twice.
- Add population group and sex to the filters.

SeanKennyNF:
Reach filter parity with serotracker.com. Resolves #339. To be merged after https://github.com/serotracker/iit-backend-v2/pull/26

Peek 2024-06-02 18-26


components/customs/filters/available-filters.tsx
```diff
@@ -336,6 +337,12 @@ export const useAvailableFilters = () => {
       label: "Serotype (DENV only)",
```
HimanshuRanka, line 337: Out of curiosity what is the purpose of this hook? It does not return anything or take any arguments so just wanted to understand.

Also see it has some Arbo only filters (and maybe sero only filters) togethere. Is that a good idea?

HimanshuRanka (approved):
approved with a question

components/customs/filters/available-filters.tsx
SeanKennyNF: 1st question: This is a hook because I want to call other hooks inside it. In this case, `useContext`. I can't call a hook in a normal function so this is a hook too :)

2nd question: Yeah the danger here is that you can specify a filter for a field that doesn't actually exist in your estimate (Ex. `populationGroup` for SarsCov2). This is helpful for stuff like `whoRegion` though since both of these can share the same filter. My ideal solution here would be that you could only use a filter if your `TEstimate` has the filterable field as a key in the object so `whoRegion` is good for both arbo and sc2 but something like `pathogen` is not. We're not there quite yet though, I've got to lab it out a bit more :man_scientist: 

Does that kind of answer the questions?

Issue #341, opened 2024-06-02
Increase percentage of data that has seroprevalence attached


SeanKennyNF:
https://3.basecamp.com/5061858/buckets/30589330/messages/7438758503#__recording_7456825783

> For seroprevalence, there shouldn’t be so many N/A values - are you calculating numerator value/denominator value? Sometimes the numerator is missing and that’s fine, you can just output seroprevalence directly from airtable.

PR #343, opened 2024-06-03
Made the modelled seroprevalence graph take into account estimates without numerators

SeanKennyNF:
Part of resolving #341. It turns out we have a lot of data without numerators so I'm calculating it myself if it's not present but seroprevalence is.

Peek 2024-06-03 10-53


HimanshuRanka:
Did we get this logic approved by the data team?

SeanKennyNF:
@HimanshuRanka, yeah that was the ask for the table and I feel like the graph and the table should be consistent. Maybe that's me making an assumption? This is what I'm basing the PR on. Does that seem fair to you too or do you want to consult with the data team?

image

HimanshuRanka:
I think getting the new data verified just once (I am sure it is ok) but to get an extra layer of approval would not hurt

PR #348, opened 2024-06-03
Fixed the country bounding box on Angola and makes the country modal appear.

### LINKED ISSUE #346: Fix the bug that makes Angola not have a country modal popup in SarsCov2Tracker.
See gif. The country modal does not pop up and applying the filter does not pan the map to Angola in the same way it does with the other countries. My guess here is that Angola does not have any data in ArboTracker and was probably never saved in our country map as a result.

Peek 2024-06-03 11-15

SeanKennyNF:
Resolves #346. Fixed the country bounding box on Angola and makes the country modal appear.

Peek 2024-06-03 11-33


lib/bounding-boxes.ts
```diff
@@ -43,7 +43,7 @@ const alphaTwoCountryCodeBoundingBoxData: Record<
   BoundingBox | undefined
 > = {
   AF: [60.53, 29.32, 75.16, 38.49],
-  A0: [11.64, -17.93, 24.08, -4.44],
+  AO: [11.64, -17.93, 24.08, -4.44],
```
SeanKennyNF, line 46: :sweat_smile:

HimanshuRanka: LOOOL

HimanshuRanka (approved):

PR #369, opened 2024-06-08
Fixed cutoff bug in the data table

### LINKED ISSUE #368: Fix cutoff in the rows of the data table.
Adding the dropdown into the ArboTracker table appears to have introduced a regression which caused. I believe this is the `overflow-x-hidden` on the cell but anything that fixes this while keeping the column the same length when the row is expanded works for me hahaha

image

SeanKennyNF:
Resolves #368

Screenshot from 2024-06-08 15-46-39
image

### DIFF
--- a/components/ui/data-table/data-table.tsx
+++ b/components/ui/data-table/data-table.tsx
@@ -86,6 +86,9 @@ export function DataTable<TData extends Record<string, unknown>, TValue>(props: 
   const table = useReactTable({
     data: props.data,
     columns: props.columns,
+    defaultColumn: {
+      size: 0
+    },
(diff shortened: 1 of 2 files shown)


components/ui/data-table/data-table.tsx
```diff
@@ -86,6 +86,9 @@ export function DataTable<TData extends Record<string, unknown>, TValue>(props:
   const table = useReactTable({
     data: props.data,
     columns: props.columns,
+    defaultColumn: {
+      size: 0
```
SeanKennyNF, line 90: Default is 150 which I knew but forgot for line 31 which was causing every column to have a width of 150 instead of being flexible.

PR #393, opened 2024-06-15
Completed list of country bounding boxes

### LINKED ISSUE #381: Complete the list of country bounding boxes.
Some country alpha two codes do not have an associated bounding box on the dashboard so the zoom does not work. Let's add bouding boxes for every single country.

(This is most noticeable with SC2Tracker which introduces a lot of new countries that didn't have arbovirus data)

SeanKennyNF:
Resolves #381. Added some countries which were not included in the library the bounding boxes were pulled from and did not have any data in Arbotracker. Mostly island nations or chains of islands.

Peek 2024-06-15 14-18


lib/bounding-boxes.ts
```diff
@@ -42,6 +42,73 @@ const alphaTwoCountryCodeBoundingBoxData: Record<
   string,
   BoundingBox | undefined
 > = {
+  AD: [1.41, 42.42, 1.79, 42.66],
```
SeanKennyNF, line 45: I'll sort the list in another PR later. I just want to include these at the top so that if you're curious which countries were included that's clear. We'll make the order pleasant later.

For chains of islands, the bounding box includes all of the islands so that's why some of these are really big. There's a couple SC2Tracker countries that weren't zoomed in on properly because they were missing from this list (Andorra, Faroe Islands, the Holy See (although that bonding box is hard to get right for some reason), Cabo Verde, etc.)

